PR #53, opened 2021-04-05
Make install script POSIX compliant

vladimyr:
Apart from being POSIX compliant this supports configuring install location through standard `PREFIX` environment variable, for example:
```sh
PREFIX="$HOME/.local" sh -c "$(curl -fsSL https://github.com/soywod/himalaya/raw/master/install.sh)"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ more flexibility.
 ## Installation
 
 ```sh
-sh -c "$(curl -sSL https://github.com/soywod/himalaya/raw/master/install.sh)"
+curl -sSL https://raw.githubusercontent.com/soywod/himalaya/master/install.sh | sh
 ```
 
 *See the [wiki section](https://github.com/soywod/himalaya/wiki/Installation)
--- a/README.md
+++ b/README.md
@@ -46,8 +46,8 @@ more flexibility.
 
 ## Installation
 
-```bash
-curl -sSL https://raw.githubusercontent.com/soywod/himalaya/master/install.sh | bash
+```sh
+sh -c "$(curl -sSL https://github.com/soywod/himalaya/raw/master/install.sh)"
 ```
 
 *See the [wiki section](https://github.com/soywod/himalaya/wiki/Installation)
--- a/install.sh
+++ b/install.sh
@@ -4,61 +4,36 @@ set -eu
 
 DESTDIR="${DESTDIR:-/}"
 PREFIX="${PREFIX:-"$DESTDIR/usr/local"}"
+RELEASES_URL="https://github.com/soywod/himalaya/releases"
 
-releases_url="https://github.com/soywod/himalaya/releases"
+system=$(uname -s | tr [:upper:] [:lower:])
 
-uname_os() {
-  os=$(uname -s | tr '[:upper:]' '[:lower:]')
+case $system in
+  msys*|mingw*|cygwin*|win*) system=windows;;
+  linux|freebsd) system=linux;;
+  darwin) system=macos;;
+  *) echo "Error: Unsupported system: $system"; exit 1;;
+esac
 
-  case $os in
-    msys*) os="windows" ;;
-    mingw*) os="windows" ;;
-    cygwin*) os="windows" ;;
-    win*) os="windows" ;;
-  esac
+if ! tmpdir=$(mktemp -d); then
+  echo "Error: Failed to create tmpdir"
+  exit 1
+else
+  trap "rm -rf $tmpdir" EXIT
+fi
 
-  echo "$os"
-}
+echo "Downloading latest $system release…"
+curl -sLo "$tmpdir/himalaya.tar.gz" "$RELEASES_URL/latest/download/himalaya-$system.tar.gz"
 
-get_system() {
-  case $(uname_os) in
-    linux | freebsd) system="linux" ;;
-    darwin) system="macos" ;;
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -45,8 +45,8 @@ more flexibility.
 
 ## Installation
 
-```bash
-curl -sSL https://raw.githubusercontent.com/soywod/himalaya/master/install.sh | bash
+```sh
+sh -c "$(curl -sSL https://github.com/soywod/himalaya/raw/master/install.sh)"
```
soywod, line 49: Why replacing `raw.githubusercontent.com` by `github.com`? In fact the cURL command can be simplified by `curl -sS https://raw.githubusercontent.com/soywod/himalaya/master/install.sh`, it prevents a redirection.

install.sh
```diff
@@ -1,32 +1,64 @@
...
+  esac
+
+  echo "$os"
+}
+
+get_system() {
+  case $(uname_os) in
+    linux | freebsd) system="linux" ;;
+    darwin) system="macos" ;;
+    windows) system="windows" ;;
```
soywod, line 27: `uname_os` and `get_system` cannot be merged together?

```suggestion
    msys*) os="windows" ;;
    mingw*) os="windows" ;;
    cygwin*) os="windows" ;;
    win*) os="windows" ;;
```

install.sh
```diff
@@ -1,32 +1,64 @@
...
+  else
+    sudo mkdir -p "$PREFIX/bin"
+    sudo install "$tmpdir/himalaya.exe" "$PREFIX/bin/himalaya"
+  fi
+
+  echo "$("$PREFIX/bin/himalaya" --version) installed!"
+}
 
-echo "$(himalaya --version) installed!"
+start
```
soywod, line 64: What's the benefit of wrapping everything in a `start` function?

soywod (changes requested):
Thank you so much for the improvement!! I just have some questions (I put them in a review).

chambln:
Sorry for the suggestions spam.  A lot of this is redundant assignments passing values around.  The gist of the program is as follows:

1. Detect OS by matching `uname -s | tr [:upper:] [:lower:]` against known values and exit nonzero if unrecognised
2. Download the corresponding release archive using curl(1)
3. Extract the archive using tar(1)
4. Copy `himalaya.exe` to `/usr/local/bin/himalaya` and make it executable
5. Clean up any temporary files upon exit using trap(1)

There may be more robust ways of doing OS detection in a POSIX friendly way, but uname(1) might be enough (what's the uname output on Windows systems?).  Could steal some some ideas from <https://github.com/dylanaraps/pfetch>.

Step 4 should probably be done as root.  Can we call it a POSIX shell script if it makes calls to sudo?  It may be better to suggest the user pipes the whole script into `sudo sh` than have sudo commands in the script.  Then the user could substitute sudo(1) for doas(1) or su(1) without having to edit the script.

install.sh
soywod: If the only benefits are for testing purposes then we can forget! I don't plan to test the `install.sh` :laughing: (except for the OS detection there is no logic). So I would remove it. [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

install.sh
```diff
@@ -1,32 +1,64 @@
...
+}
+
+start() {
+  system=$(get_system)
+  if [ -z "$system" ]; then
+    echo "Error: Unsupported system: $system"
+    exit 1
+  fi
+
+  if ! tmpdir=$(mktemp -d); then
```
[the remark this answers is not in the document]
soywod: I don't mind if it's not POSIX, at least it's better than what we got before (I think). Feel free to propose any better alternative!

install.sh
```diff
@@ -1,32 +1,64 @@
...
+
+get_system() {
+  case $(uname_os) in
+    linux | freebsd) system="linux" ;;
+    darwin) system="macos" ;;
+    windows) system="windows" ;;
+  esac
+
+  echo "$system"
+}
```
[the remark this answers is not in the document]
soywod: It's way more concise this way, I approve also!

install.sh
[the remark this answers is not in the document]
chambln:
```sh
trap 'rm -f /tmp/himalaya.tar.gz /tmp/himalaya.exe' EXIT
mkdir -p /tmp
...
```

soywod:
@vladimyr what do you think about @chambln propositions?

soywod:
@chambln I did all the changes you proposed. Could you check one more time before I merge?

vladimyr:
Sorry for being unresponsive, I've had limited FOSS time for the last few days. Anyway if you are still interested I'll gladly go through your questions and explain the rationale behind my choices.

soywod:
@vladimyr I am sorry, I took the initiative to push commits on your branch to merge faster. Even if the discussion is not terminated with @chambln, the result is still better than what we got before :smiley:

> Anyway if you are still interested I'll gladly go through your questions and explain the rationale behind my choices.

With great pleasure! Thanks for you contribution anyway.

chambln:
Looks much better already!

If POSIX compliance is the goal then sudo must not be used in the script. I think it's wise to leave privilege elevation to the person running the script anyway. We can still *suggest* using sudo in `README.md`, i.e.

```sh
curl ... | sudo sh
```

PR #62, opened 2021-04-05
Look for fzf instead of fzf.vim

AckslD:
Since `fzf` is what is required and not necessarily `fzf.vim` it would be better to just check for this in the `rtp`.

### DIFF
--- a/vim/autoload/himalaya/mbox.vim
+++ b/vim/autoload/himalaya/mbox.vim
@@ -27,7 +27,7 @@ endfunction
 function! himalaya#mbox#input()
   try
     let mboxes = map(s:cli("mailboxes", [], "Fetching mailboxes"), "v:val.name")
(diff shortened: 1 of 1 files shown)


soywod:
In fact `fzf.vim` is needed to call the function `fzf#run`, so I think we should keep the `rtp` as it is.

AckslD:
@soywod I think it's actually defined in `fzf`, see [here](https://github.com/junegunn/fzf/blob/master/plugin/fzf.vim#L447). I only have `fzf` installed and not `fzf.vim` and with this change things to work for me :)

soywod:
Oh yes you right, I'm sorry! You totally right. I tried and seems to work also for me, so I merge.

soywod:
Thank you :tada:

AckslD:
No problem :) Cool!

PR #66, opened 2021-04-06
Make install.sh more robust and portable

chambln:
With these changes, Bash is no longer required to run the installer—any POSIX compliant shell will do.

Seems like shells other than Bash don't necessarily provide `$OSTYPE` and so the OS detection logic has been adapted to use uname(1) instead.  Ideally needs testing on Windows, macos and supported BSDs before being merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,10 @@ more flexibility.
 ## Installation
 
 ```bash
-curl -sSL https://raw.githubusercontent.com/soywod/himalaya/master/install.sh | bash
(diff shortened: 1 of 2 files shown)


chambln:
After all that, I only just noticed PR #53 :facepalm:

soywod:
> After all that, I only just noticed PR #53

I'm sorry for you dude :laughing: I close this PR let's continue the discussion on the existing one #53. Feel free to propose your changes there!

PR #67, opened 2021-04-06
Remove unused clap depdenencies

alexwennerberg:
I believe most of the default clap dependencies are used, this shrinks the dependency tree slightly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ authors = ["soywod <clement.douin@posteo.net>"]
 edition = "2018"
 
 [dependencies]
-clap = "2.33.3"
+clap = {version = "2.33.3", default-features = false, features=["suggestions"]}
 error-chain = "0.12.4"
 imap = "2.4.0"
 lettre = "0.10.0-beta.3"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -6,7 +6,7 @@ authors = ["soywod <clement.douin@posteo.net>"]
 edition = "2018"
 
 [dependencies]
-clap = "2.33.3"
+clap = {version = "2.33.3", default-features = false, features=["suggestions"]}
```
soywod, line 9: By doing so, you also remove colors in all help menus since it's part of the [default features](https://github.com/clap-rs/clap#features-enabled-by-default). I think we should put back the "color" feature at least. For now I manage colors myself, which is sth to avoid. I should use the same [libs](https://github.com/clap-rs/clap/blob/8b721a2bcdcf412c67342a6bdf2d540150dbd37e/Cargo.toml#L92) as clap.

soywod (changes requested):
Thanks for this optimization!

Cargo.toml
alexwennerberg: Added!

PR #68, opened 2021-04-06
else if -> elseif

AckslD:
otherwise gives:
```
Vim(else):E488: Trailing characters:     else if &rtp =~
```


soywod:
Thanks for pointing out this typo :man_facepalming: :laughing:

PR #69, opened 2021-04-06
Preview mailboxes with telescope

AckslD:
I started adding preview of mailboxes using telescope. It works well, see (blurred) screencapture below. The only issue is when a mailbox is empty since there is then an error somewhere in the vimscripts I think. @soywod Is there a way to capture this error or perhaps check beforehand if the mailbox is empty or not?

himalaya

### DIFF
--- a/vim/autoload/himalaya/mbox.lua
+++ b/vim/autoload/himalaya/mbox.lua
@@ -18,11 +18,13 @@ function mbox_picker(mboxes)
                 preview_command = function(entry, bufnr)
                     vim.api.nvim_buf_call(bufnr, function()
                         local page = 0 -- page 0 for preview
-                        local success, output = pcall(vim.fn['himalaya#msg#list_preview'], entry.value, page)
+                        local success, output = pcall(vim.fn['himalaya#msg#list_with'], entry.value, page)
                         if not (success) then
                             vim.cmd('redraw')
                             vim.bo.modifiable = true
-                            vim.api.nvim_buf_set_lines(bufnr, 0, -1, true, {"Error: "..output})
+                            local errors = vim.fn.split(output, '\n')
+                            errors[1] = "Errors: "..errors[1]
+                            vim.api.nvim_buf_set_lines(bufnr, 0, -1, true, errors)
                         end
                     end)
                 end
--- a/vim/autoload/himalaya/mbox.lua
+++ b/vim/autoload/himalaya/mbox.lua
@@ -18,7 +18,7 @@ function mbox_picker(mboxes)
                 preview_command = function(entry, bufnr)
                     vim.api.nvim_buf_call(bufnr, function()
                         local page = 0 -- page 0 for preview
-                        local success, output = pcall(vim.fn['himalaya#msg#list_with'], entry.value, page)
+                        local success, output = pcall(vim.fn['himalaya#msg#list_with'], entry.value, page, true)
                         if not (success) then
                             vim.cmd('redraw')
                             vim.bo.modifiable = true
--- a/vim/autoload/himalaya/mbox.lua
+++ b/vim/autoload/himalaya/mbox.lua
@@ -17,10 +17,11 @@ function mbox_picker(mboxes)
                 ordinal = entry,
                 preview_command = function(entry, bufnr)
                     vim.api.nvim_buf_call(bufnr, function()
-                        local success, _ = pcall(vim.fn['himalaya#mbox#post_input'](entry.value))
+                        local success, output = pcall(vim.fn['himalaya#mbox#post_input'], entry.value)
                         if not (success) then
                             -- TODO does not work since buffer is not modifiable
-                            -- vim.api.nvim_buf_set_lines(bufnr, 0, -1, true, {"Empty mailbox"})
+                            vim.bo.modifiable = true
+                            vim.api.nvim_buf_set_lines(bufnr, 0, -1, true, {"Error: "..output})
                         end
                     end)
                 end
--- a/vim/autoload/himalaya/mbox.lua
+++ b/vim/autoload/himalaya/mbox.lua
@@ -3,11 +3,30 @@ local action_state = require('telescope.actions.state')
 local finders = require('telescope.finders')
 local pickers = require('telescope.pickers')
 local sorters = require('telescope.sorters')
+local previewers = require('telescope.previewers')
 
 function mbox_picker(mboxes)
   pickers.new {
     results_title = 'Mailboxes',
-    finder = finders.new_table(mboxes),
+    finder = finders.new_table {
+        results = mboxes,
+        entry_maker = function(entry)
+            return {
+                value = entry,
+                display = entry,
+                ordinal = entry,
+                preview_command = function(entry, bufnr)
+                    vim.api.nvim_buf_call(bufnr, function()
+                        local success, _ = pcall(vim.fn['himalaya#mbox#post_input'](entry.value))
+                        if not (success) then
+                            -- TODO does not work since buffer is not modifiable
+                            -- vim.api.nvim_buf_set_lines(bufnr, 0, -1, true, {"Empty mailbox"})
+                        end
+                    end)
+                end
+            }
+      end,
+    },
     sorter = sorters.fuzzy_with_index_bias(),
     attach_mappings = function(prompt_bufnr)
       actions.select_default:replace(function()
@@ -18,5 +37,6 @@ function mbox_picker(mboxes)
 
       return true
     end,
+    previewer = previewers.display_content.new({})
   }:find()
 end
--- a/vim/autoload/himalaya/mbox.lua
+++ b/vim/autoload/himalaya/mbox.lua
@@ -17,9 +17,10 @@ function mbox_picker(mboxes)
                 ordinal = entry,
                 preview_command = function(entry, bufnr)
                     vim.api.nvim_buf_call(bufnr, function()
-                        local success, output = pcall(vim.fn['himalaya#mbox#post_input'], entry.value)
+                        local page = 0 -- page 0 for preview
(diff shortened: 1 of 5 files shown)


vim/autoload/himalaya/mbox.lua
```diff
@@ -3,11 +3,30 @@ local action_state = require('telescope.actions.state')
...
+        entry_maker = function(entry)
+            return {
+                value = entry,
+                display = entry,
+                ordinal = entry,
+                preview_command = function(entry, bufnr)
+                    vim.api.nvim_buf_call(bufnr, function()
+                        local success, _ = pcall(vim.fn['himalaya#mbox#post_input'](entry.value))
+                        if not (success) then
+                            -- TODO does not work since buffer is not modifiable
```
soywod, line 22: Seems to be the correct approach for me. Maybe you can change the `modifiable` this way: `vim.bo.modifiable = true` ([sources](https://github.com/nanotee/nvim-lua-guide#using-meta-accessors)), and replace the buffer with the error returned by `himalaya#mbox#post_input` (or whatever).

AckslD: Good point @soywod! I also noticed that I had made a mistake of actually calling the function instead of giving it to `pcall` so fixed that :)

However, it doesn't seem to really work. Not sure if errors are properly propagated up to lua here. I.e. the error is created by calling `echoerr` I assume which perhaps does not get treated as an error by `pcall` but simply by vim.

soywod: Indeed, all top-level functions are wrapped in a `try` block
https://github.com/soywod/himalaya/blob/0f1c490dad7f1df1b2cad5091ccdb8d66b26eb25/vim/autoload/himalaya/msg.vim#L21-L41
What can be done is to call directly the CLI from lua, either by calling the vim fn `himalaya#shared#cli#call` or by spawning a command. What do you think?

AckslD: @soywod Would it be okay to throw the error in the underlying vimscript-function? See latest commit. This makes the preview work but not sure if it breaks something else. What's the reason for catching any error?

If this does not work we could indeed directly call the cli, however it would give some duplicated code in the vimscript and lua files I guess. Which is maybe fine but would be nice to avoid.

vim/autoload/himalaya/shared/log.vim
```diff
@@ -7,4 +7,5 @@ function! himalaya#shared#log#err(msg)
   echohl ErrorMsg
   echomsg a:msg
   echohl None
+  throw a:msg
```
soywod, line 10: There is one place in the code where this function is called many times in a row: https://github.com/soywod/himalaya/blob/14d8a3fecdf34aaf0d554ed08b60e48ed019e98b/vim/autoload/himalaya/shared/cli.vim#L14-L17

So I am not sure it's the best approach.

soywod (changes requested):

vim/autoload/himalaya/mbox.lua
```diff
@@ -3,11 +3,31 @@ local action_state = require('telescope.actions.state')
...
+    finder = finders.new_table {
+        results = mboxes,
+        entry_maker = function(entry)
+            return {
+                value = entry,
+                display = entry,
+                ordinal = entry,
+                preview_command = function(entry, bufnr)
+                    vim.api.nvim_buf_call(bufnr, function()
+                        local success, output = pcall(vim.fn['himalaya#mbox#post_input'], entry.value)
```
soywod, line 20: There is an other problem. `himalaya#mbox#post_input` modifies the current mailbox and the current page: https://github.com/soywod/himalaya/blob/14d8a3fecdf34aaf0d554ed08b60e48ed019e98b/vim/autoload/himalaya/mbox.vim#L56-L58

I am afraid we will need an other function to preview without changing the cursors!

vim/autoload/himalaya/shared/log.vim
AckslD: I see! What would you think of extracting out the main body of `himalaya#msg#list` into a seperate function which does not catch the error and we can call from the preview and from `himalaya#msg#list` catching any error? I.e. something like (or maybe named differently if you don't want to rename the original function)
```vim
function! himalaya#msg#list()
  let mbox = himalaya#mbox#curr_mbox()
  let page = himalaya#mbox#curr_page()
  let msgs = s:cli("--mailbox %s list --page %d", [shellescape(mbox), page], printf("Fetching %s messages", mbox))
  let msgs = map(msgs, "s:format_msg_for_list(v:val)")
  let buftype = stridx(bufname("%"), "Himalaya messages") == 0 ? "file" : "edit"
  execute printf("silent! %s Himalaya messages [%s] [page %d]", buftype, mbox, page + 1)
  setlocal modifiable
  silent execute "%d"
  call append(0, s:render("list", msgs))
  silent execute "$d"
  setlocal filetype=himalaya-msg-list
  let &modified = 0
  execute 0
endfunction

function! himalaya#msg#try_list()
  try
    call himalaya#msg#list()
  catch
    if !empty(v:exception)
      redraw | call himalaya#shared#log#err(v:exception)
    endif
  endtry
endfunction
```

vim/autoload/himalaya/mbox.lua
AckslD: I see! Could `curr_mbox`/`curr_page` be given as arguments to `himalaya#mbox#post_input` instead of being saved as script variables?

vim/autoload/himalaya/shared/log.vim
soywod: Yes I like the idea, this way you can reuse the same function inside lua!

vim/autoload/himalaya/mbox.lua
soywod: We need an other function anyway, since we have to call inside the new function `himalaya#msg#list` (the tryless one from your proposition https://github.com/soywod/himalaya/pull/69#discussion_r609513786). Maybe you can rename `himalaya#msg#list` to `himalaya#msg#list_preview`, and give as parameter the `mbox`? The page can stay at 0 since it's just a preview.

AckslD:
@soywod I updated the himalaya#msg#list-function now, let me know what you think. One thing to point out is that I added a `redraw` call in the preview function to avoid the `Press ENTER or type command to continue`-message when there is an error. The message `Fetching <mailbox> messages [<status>]`-message will then not be showed in case of error but instead printed in the preview buffer.

I also change how errors are thrown in `himalaya#shared#cli#call` in order to get the error message in the preview buffer.

vim/autoload/himalaya/msg.vim
```diff
@@ -18,21 +18,25 @@ function! s:format_msg_for_list(msg)
...
-    let buftype = stridx(bufname("%"), "Himalaya messages") == 0 ? "file" : "edit"
-    execute printf("silent! %s Himalaya messages [%s] [page %d]", buftype, mbox, page + 1)
-    setlocal modifiable
-    silent execute "%d"
-    call append(0, s:render("list", msgs))
-    silent execute "$d"
-    setlocal filetype=himalaya-msg-list
-    let &modified = 0
-    execute 0
+    call himalaya#msg#list_preview(mbox, page)
```
soywod, line 39: I did not know you wanted to call the `list_preview` fn here also, and in fact it makes sense. This way you reuse the logic. But the name does not match anymore. Maybe sth like `list_with`? Or whatever you find more appropriate ;)

vim/autoload/himalaya/shared/cli.vim
```diff
@@ -14,8 +14,8 @@ function! himalaya#shared#cli#call(cmd, args, log)
       redraw | call himalaya#shared#log#info(printf("%s [ERR]", a:log))
       for line in split(res, "\n")
         call himalaya#shared#log#err(line)
+        throw line
```
soywod, line 17: The CLI returns a list a errors (a kind of error chain) separated by `\n`. By throwing there we will get only the first error, which is not good. In fact we want this piece of code to throw only when it's called from the telescope previewer, right? Maybe we can pass a new parameter to the function `himalaya#shared#cli#call`, sth like `should_throw` or `throwable`. Feel free to propose a better idea if you have!

soywod (changes requested):
I think we are almost good :) thanks for your awesome work!

vim/autoload/himalaya/msg.vim
AckslD: Good point, changes to `list_with`

vim/autoload/himalaya/shared/cli.vim
AckslD: I was thinking of that before but wasn't sure if the function should throw an error as it did before as well? It earlier did `throw ""` after the for-loop. I now added that back and made it `throw res` instead which we can capture in the preview.

soywod: What I had in mind is more sth like:

```viml
function! himalaya#shared#cli#call(cmd, args, log, should_throw)
...

for line in split(res, "\n")
  call himalaya#shared#log#err(line)
endfor

if a:should_throw
  throw ""
endif
```

This way, only the `list_preview` will throw:

```viml
function! himalaya#msg#list_preview(mbox, page)
  let msgs = s:cli("--mailbox %s list --page %d", [shellescape(a:mbox), a:page], printf("Fetching %s messages", a:mbox), 1)
```

AckslD: Added the `should_throw` argument 👍

soywod:
Looks good to me, I guess we can merge. Thanks for your contribution!

AckslD:
Cool 🎉 I'm happy that I could contribute :)

soywod:
I opened an issue related to this feature: https://github.com/soywod/himalaya/issues/92. Feel free to contribute if you want (since you already worked on this part) :smiley:.

PR #73, opened 2021-04-07
Add support for installing using Nix

srid:
[What's Nix?](https://serokell.io/blog/what-is-nix) It is the best way to [install](https://neuron.zettel.page/install) software in a *reproducible* and automated fashion!

Resolves #70 

- [x] Running `nix-build` produces a binary at `./result/`
- [x] Consider [flake](https://www.tweag.io/blog/2020-05-25-flakes/) support
- [x] Nix best practices
    - [x] Respect .gitignore
    - [x] Pin nixpkgs
- [x] Documentation
  - [x] Document Nix maintenance workflow
      - how to keep version in sync with Cargo.toml (automate?)
      - how to update the Cargo dependency sha (done manually when Cargo deps change)
  - [x] Add nix instructions to README

PR #76, opened 2021-04-08
Add support for installing using Nix (zero maintenance)

### LINKED ISSUE #70: Nix support
Adding a `default.nix` would help users on NixOS, or using Nix, to easily build himalaya. Here's the default.nix that I used to build the current master branch:

```nix
let 
  pkgs = import <nixpkgs> {};
in 
pkgs.rustPlatform.buildRustPackage rec {
  pname = "himalaya";
  version = "0.2.2";

  src = ./.;

  nativeBuildInputs = with pkgs; [
    pkg-config 
    openssl.dev
  ];
  PKG_CONFIG_PATH = "${pkgs.openssl.dev}/lib/pkgconfig";

  cargoSha256 = "sha256-i8YKE5Z2xoNyjFktRRVA45zD6Qa0aPNC0AOjHUIglgk=";

  meta = with pkgs.stdenv.lib; {
    description = "Minimalist CLI email client";
    homepage = "https://github.com/soywod/himalaya";
  };
}
```

srid:
This is an improved version of #73 (and resolves #70) wherein the need to change the version and hash every time the Cargo files change is avoided. Now you can just forget about the .nix files (unless a new C dependency is being introduced).


flake.nix
```diff
@@ -0,0 +1,72 @@
...
+       let 
+          name = "himalaya";
+          pkgs = import nixpkgs { 
+            inherit system; 
+            overlays = [ 
+              rust-overlay.overlay
+              (self: super: {
+                # Because rust-overlay bundles multiple rust packages into one
+                # derivation, specify that mega-bundle here, so that naersk
+                # will use them automatically.
```
balsoft, line 30: This is not using naersk anymore

flake.nix
```diff
@@ -0,0 +1,72 @@
...
+            # Individual crate overrides go here
+            # Example: https://github.com/balsoft/simple-osd-daemons/blob/6f85144934c0c1382c7a4d3a2bbb80106776e270/flake.nix#L28-L50
+          };
+          nativeBuildInputs = with pkgs; [ rustc cargo pkgconfig openssl.dev ];
+          buildEnvVars = {
+            PKG_CONFIG_PATH = "${pkgs.openssl.dev}/lib/pkgconfig";
+          };
+          rootCrateBuild = pkgs.lib.overrideDerivation project.rootCrate.build (oldAttrs: {
+            inherit nativeBuildInputs;
+          } // buildEnvVars);
```
balsoft, line 51: I think this belongs in the "Individual crate overrides go here" part, as `himalaya = oldAttrs: <...>`

flake.nix
```diff
@@ -0,0 +1,72 @@
...
+          packages.${name} = rootCrateBuild;
+
+          # `nix build`
+          defaultPackage = packages.${name};
+
+          # `nix run`
+          apps.${name} = utils.lib.mkApp {
+            inherit name;
+            drv = packages.${name};
+          };
```
balsoft, line 62: BTW this is not really necessary; `nix run .#$NAME` by default will build `packages.${system}.$NAME` and run `$out/bin/$NAME` from that package.

flake.nix
```diff
@@ -0,0 +1,72 @@
...
+          project = pkgs.callPackage (generatedCargoNix {
+            inherit name;
+            src = ./.;
+          }) {
+            # Individual crate overrides go here
+            # Example: https://github.com/balsoft/simple-osd-daemons/blob/6f85144934c0c1382c7a4d3a2bbb80106776e270/flake.nix#L28-L50
+          };
+          nativeBuildInputs = with pkgs; [ rustc cargo pkgconfig openssl.dev ];
+          buildEnvVars = {
+            PKG_CONFIG_PATH = "${pkgs.openssl.dev}/lib/pkgconfig";
```
balsoft, line 47: Why not just put `openssl.dev` into `buildInputs` of the package (as opposed to `nativeBuildInputs`)?

flake.nix
```diff
@@ -0,0 +1,72 @@
...
+          inherit (import "${crate2nix}/tools.nix" { inherit pkgs; })
+            generatedCargoNix;
+          project = pkgs.callPackage (generatedCargoNix {
+            inherit name;
+            src = ./.;
+          }) {
+            # Individual crate overrides go here
+            # Example: https://github.com/balsoft/simple-osd-daemons/blob/6f85144934c0c1382c7a4d3a2bbb80106776e270/flake.nix#L28-L50
+          };
+          nativeBuildInputs = with pkgs; [ rustc cargo pkgconfig openssl.dev ];
```
balsoft, line 45: `openssl.dev` is definitely doesn't belong in `nativeBuildInputs`, it should go in `buildInputs`. `nativeBuildInputs` is for packages that provide binaries for the build process, `buildInputs` is for libraries.

balsoft (commented):
Some nitpicks, looks great otherwise!

BTW, maybe you could consider adding `.envrc` with `use flake` for those of us who use `direnv+nix-direnv`.

flake.nix
srid: Right, I'll remove it. Do you think this "hack" is the correct approach though?

flake.nix
srid: Removing those lines gives,

```
error: undefined variable 'apps'

       at /nix/store/ri3ylxi49k9ym030q6c7syydkcdxvd0v-source/flake.nix:59:24:

           58|           # `nix run`
           59|           defaultApp = apps.${name};
             |                        ^
           60|

```

What would you set the `defaultApp` value too? If I remove that as well, I get:

```
❯ nix run
warning: Git tree '/home/srid/external/himalaya' is dirty
error: unable to execute '/nix/store/hzciv1n3rryqkvwaikfq7v22cmclfph1-rust_himalaya-0.2.3/bin/rust_himalaya': No such file or directory
```

Which kind of makes sense as the default `nix build` produces a derivation named differently.

flake.nix
balsoft: It's alright, but note that rust-overlay currently doesn't play too well with newer Nix versions, because of changes to how fixed-hash derivations are handled. In particular, this happens if you try to build on nix 2.4pre without first building on older Nix:

```
error: hash mismatch in fixed-output derivation '/nix/store/az96g89ncj19m2mqgnic7ikkmm0bcgfh-cargo-1.51.0-x86_64-unknown-linux-gnu.tar.xz.drv':
         specified: sha256-LFV+RIwUXtdzuq4NZTNEmUe7Ew+PiioIdtCPVbdKMT8=
            got:    sha256-zx+SwH7GK9W0E9BMbYiouohc36OoYOnmZ/eHS3/u4BI=
```

flake.nix
balsoft: Oh, right, sorry. Didn't realise crate2nix added a `rust_` prefix to the name. Keep `apps` and `defaultApp` as-is, then.

flake.nix
srid: I remember moving it to `nativeBuildInputs` just to get the openssl stuff to build. Moved it back, now.

flake.nix
srid: Wait, so this would be broken in a fresh machine with Flakes enabled (as that will use 2.4pre)?

balsoft: Yep, unfortunately so. You can add a note about it somewhere in the readme. It can be fixed with `nix shell nixpkgs#nix -c nix-build`

flake.nix
balsoft: I mean this shouldn't be needed anymore with `openssl` in `buildInputs`

srid:
@soywod This PR fails for 0.2.4; I believe that's because `Cargo.lock` wasn't updated along with the commit that changed `Cargo.toml`.

Error message,

```
++ crate2nix generate -f ./Cargo.toml -o Cargo-generated.nix -h /nix/store/i11yfpm336cwlzl3idsxiq6dm067v3cs-himalaya-crate2nix/crate-hashes.json
Error: while retrieving metadata about ./Cargo.toml: Error during execution of `cargo metadata`: error: the lock file /build/vv6c95617dy8c6hxl0qvh32crl7jqpyj-source/Cargo.lock needs to be updated but --locked was passed to prevent this
```

soywod:
Ah yes indeed, I bumped the version without launching back cargo build, I will update tomorrow

srid:
> @soywod Since I don't know enough about Nix, I plan to dive in this weekend to understand at least the concept. This way I will understand better what you did and I will be able to merge your work.

By the way, Nix is also useful for non-development purposes, such as wrapping `himalaya` in conjunction with other tools (pandoc, bat, fzf, etc.) to provide extra workflows, like this one for reading email in CLI:

https://github.com/srid/nixos-config/commit/6362b4d9b4c63dab8e156e918969e1da9aedbb3d

flake.nix
balsoft: @srid himalaya compiles correctly without this line now, so maybe remove it?

srid:
@balsoft General nix flakes question: is there a way to configure the himalaya flake.nix so that it can be made to accept home-manager style config params? That is, to specify the values in `~/.config/himalaya/config.toml` (but without having to create this file manually)?

balsoft:
Well, you can export a NixOS module that uses home-manager, or a home-manager module (though that's not as standardized).

srid:
I've created a template repo based on this work: https://github.com/srid/rust-nix-template

srid:
@balsoft Do you know what the idiomatic way is to add dev-shell Rust dependencies in Nix?

For eg., to make `cargo watch` available in nix-shell: https://github.com/passcod/cargo-watch

Or certain crate tools available in nix-shell, eg: `trunk` from https://yew.rs/docs/en/next/getting-started/project-setup/using-trunk

flake.nix
```diff
@@ -0,0 +1,80 @@
...
+          # `nix run`
+          apps.${name} = utils.lib.mkApp {
+            inherit name;
+            drv = packages.${name};
+          };
+          defaultApp = apps.${name};
+
+          # `nix develop`
+          devShell = pkgs.mkShell {
+            inherit buildInputs nativeBuildInputs;
```
balsoft, line 75:
```suggestion
            inputsFrom = builtins.attrValues self.packages.${system};
            buildInputs = with pkgs; [ cargo cargo-watch trunk ];
```

srid:
@balsoft Do you know of an idiomatic way to override a crate to point to a git fork? 

I noticed that Cargo.toml supports this, but we can't use it because the fetch will fail during nix build (due to blocked network access):

```toml
# Cargo.toml
postgrest = { git = "https://github.com/srid/postgrest-rs", branch = "notokio" }
```

I also tried adding something like this to `defaultCrateOverrides`, but that doesn't work for some reason (build is still using the crate registry's package stored in nix store),

```nix
                postgrest = oldAttrs: with pkgs; {
                  src = inputs.postgrest-rs; 
                };
 ```

srid:
Ah. Looks like you have to specifying the source in *both* places (Cargo.toml and flake input+override); then it works ... mysteriously.

balsoft:
> the fetch will fail during nix build (due to blocked network access):

I thought I fixed this here: https://github.com/kolloch/crate2nix/pull/154 (but only on Nix 2.4)

srid:
@soywod I believe this PR is ready to be merged anytime. 

(I keep merging it with `master` only because I use this branch directly in my nixos configuration).

soywod:
Thanks a lot for your giant contribution! Feel free to write an entry in the wiki about Nix (if you judge it necessary, ofc).

soywod:
In fact, maybe it makes sense to move the Nix section from the README to the wiki, in the [Installation](https://github.com/soywod/himalaya/wiki/Installation) section?

PR #99, opened 2021-04-13
add completion subcommands

remche:
This PR allows to generate Bash, Zsh and Fish completions with `himalaya bash-completions`, `himalaya zsh-completions`, `himalaya fish-completions`.

This is my very first Rust code... Please be indulgent :grin:


soywod:
Thank you so much for you contribution! I was thinking about adding completion, but you were faster than me :laughing:. Your PR is very good, I just pushed a commit to:

- Rename `complete` to `completion`
- Remove the `unwrap` (make use of [`error_chain`]())
- Add an entry in the README (I will probably move it to the wiki later)

README.md
```diff
@@ -237,6 +238,32 @@ systemctl --user enable himalaya.service
...
+
+```sh
+# For bash shells
+himalaya bash-completions
+
+# For zsh shells
+himalaya zsh-completions
+
+# For fish shells
+himalaya fish-completions
```
srid, line 251: Perhaps these should be `himalaya completions {bash, fish, zsh}` to be consistent with the rest of the subcommand structure (`himalaya template ...` for instance).

remche: Agree.

soywod: I also agree, it's a good point. I made the change in the last commit (https://github.com/soywod/himalaya/commit/aad74add5296ceadda2b1be3d4ad778a51f6ecd4). I also redefine shells as arguments instead of subcommands for the completion + help.

Issue #84, opened 2021-04-09
Ignore self-signed certificate option


remche:
Enabling self-signed certificates (and/or non-encrypted connections) would allow to play with test servers like [GreenMail](https://github.com/greenmail-mail-test/greenmail) (and write some CI functional tests ?).

soywod:
I'm not sure this is related to Himalaya. A self-signed cert can be trusted from the OS side, like in [this article](https://iptq.io/setup/command-line-email-with-aerc/#setting-up-aerc-with-protonmail):

```sh
sudo trust anchor --store self-signed.crt
```

ibotty:
I don't know how to use `trust` when I am not system admin.  It would be great to allow that use case.

soywod:
We could attach a new property to the config but it requires some work. This is clearly not a priority for the `v1`. I would advise to make your system trust the certs for now. Feel free to contribute if you need this feature :wink:

remche:
My use case is mainly for testing with greenmail which generates a new cert at each run.
I might take the opportunity to learn rust ;)

soywod:
With great pleasure :wink: feel free to use the [discussion feature](https://github.com/soywod/himalaya/discussions) if needed!

PR #103, opened 2021-04-14
allow insecure TLS option

remche:
This PR allow to use insecure connection for imap and tls via `imap-insecure` and `smtp-insecure` as discussed in #84.
My use case is for testing with [GreenMail](https://greenmail-mail-test.github.io/greenmail/).

### DIFF
--- a/src/smtp.rs
+++ b/src/smtp.rs
@@ -1,5 +1,12 @@
 use error_chain::error_chain;
-use lettre::{self, transport::smtp::SmtpTransport, Transport};
+use lettre::{self, 
+    transport::{
+        smtp::SmtpTransport,
+        smtp::client::Tls,
+        smtp::client::TlsParameters
+    },
+    Transport
+};
 
 use crate::config::model::Account;
 
@@ -19,8 +26,15 @@ pub fn send(account: &Account, msg: &lettre::Message) -> Result<()> {
         SmtpTransport::relay
     };
 
+    let tls = TlsParameters::builder(account.smtp_host.to_string())
+        .dangerous_accept_invalid_hostnames(account.smtp_insecure())
+        .dangerous_accept_invalid_certs(account.smtp_insecure())
+        .build()
+        .unwrap();
+
     smtp_relay(&account.smtp_host)?
         .port(account.smtp_port)
+        .tls(Tls::Wrapper(tls))
         .credentials(account.smtp_creds()?)
         .build()
         .send(msg)?;
--- a/src/config/model.rs
+++ b/src/config/model.rs
@@ -25,12 +25,14 @@ pub struct Account {
     pub imap_host: String,
     pub imap_port: u16,
     pub imap_starttls: Option<bool>,
+    pub imap_insecure: Option<bool>,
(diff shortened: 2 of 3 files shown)


soywod:
Your PR looks perfect to me! I can merge straight. Could you also update the [Configuration wiki section](https://github.com/soywod/himalaya/wiki/Configuration)?

src/smtp.rs
```diff
@@ -19,8 +26,15 @@ pub fn send(account: &Account, msg: &lettre::Message) -> Result<()> {
...
 
+    let tls = TlsParameters::builder(account.smtp_host.to_string())
+        .dangerous_accept_invalid_hostnames(account.smtp_insecure())
+        .dangerous_accept_invalid_certs(account.smtp_insecure())
+        .build()
+        .unwrap();
+
     smtp_relay(&account.smtp_host)?
         .port(account.smtp_port)
+        .tls(Tls::Wrapper(tls))
```
paolobarbolini, line 37: I just saw this PR, but forcing `Wrapper` means that even when starttls is turned on, TLS will be used anyway

soywod: But `STARTTLS` is just a `TLS` extension, so theoretically the `.tls(...)` method may affect only the `TLS` part without interfering with the `STARTTLS` part? I guess integration tests need to be set up to catch all those tricky behaviours.

paolobarbolini: The [`relay`](https://docs.rs/lettre/0.10.0-beta.3/src/lettre/transport/smtp/transport.rs.html#51-57) and [`starttls_relay`](https://docs.rs/lettre/0.10.0-beta.3/src/lettre/transport/smtp/transport.rs.html#71-77) methods are just convenience functions for setting the right variant of the [`Tls`](https://docs.rs/lettre/0.10.0-beta.3/lettre/transport/smtp/client/enum.Tls.html) enum.

`Tls::Required` = The server either supports `STARTTLS` and allows lettre to upgrade the initial plain text connection to TLS or we fail.
`Tls::Wrapped` = We open the connection directly with TLS.

remche: I'm a Rust/TLS noob but I did not see any issue with starttls on the few tests I drove.

soywod: I didn't realized you are part of the lettre core team @paolobarbolini! Thanks for your precious help. So it means that the `insecure` options are not compatible with `STARTTLS`? I mean, there is no way to enable `STARTTLS` AND pass a custom TLS builder with `dangerous_accept_*` options?

paolobarbolini: > So it means that there is no way to enable `STARTTLS` AND pass a custom TLS builder with `dangerous_accept_*` options?

It's totally possible. The change that needs to be done is passing `.tls(...)` the `Tls::Required(tls)` variant instead of the `Tls::Wrapper(tls)` variant when `account.smtp_starttls()` is `true`.

soywod: Awesome! @remche would you like to take care of this?

remche: Will do asap

PR #105, opened 2021-04-16
allow insecure TLS option, second try

remche:
change for #103 after @paolobarbolini explanation and help !


soywod:
Thanks! :tada:

PR #107, opened 2021-04-18
Improve portability of install.sh

chambln:
Read the commit messages for commentary. See also #53 and #66.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -24,7 +24,8 @@ tmpdir=$(mktemp -d) || die "Failed to create tmpdir"
 trap "rm -rf $tmpdir" EXIT
 
 echo "Downloading latest $system release…"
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -2,6 +2,11 @@
 
 set -eu
 
+die() {
+    printf '%s\n' "$1" >&2
+    exit "${2-1}"
```
soywod, line 7: Could you explain this line?

soywod:
Thank you! :tada:

install.sh
chambln: Usage is

 ```sh
die MESSAGE [EXIT_STATUS]
```

`>&2` means print to stderr rather than stdout.  `exit ${2-1}` means if `EXIT_STATUS` was not provided, default to `1`.

Issue #72, opened 2021-04-07
Cannot login to IMAP server


nbouliol:
Hi !

I have this error :

```bash
❯ himalaya
Cannot login to IMAP server
 ↳ No Response: Empty username or password.
```

And this is my config file :

```toml
name = "Nicolas Bouliol"
downloads-dir = "~/himalaya"
signature = "Regards,"


[gmail]
default = true
email = "cheuveux@gmail.com"

imap-host = "imap.gmail.com"
imap-port = 993
imap-login = "cheuveux@gmail.com"
imap-passwd-cmd = ""

smtp-host = "smtp.gmail.com"
smtp-port = 487
smtp-login = "cheuveux@gmail.com"
smtp-passwd-cmd = ""
```

`imap-passwd-cmd` and `smtp-passwd-cmd` are generated passwords as mentionend in #52 

What did I miss ?

soywod:
`imap-passwd-cmd` and `smtp-passwd-cmd` need to be shell commands, not directly the password. For testing purposes you can put `echo "your-token"`, but the best is to use a password manager like [`pass`](https://www.passwordstore.org/) so you can put `pass show gmail`. Everything is explained in the [wiki](https://github.com/soywod/himalaya/wiki/Configuration). Let me know if it fixes your problem.

nbouliol:
Ok my bad, thanks for your response

chawyehsu:
Only giving `cmd` way to configure passwd is difficult for Windows users. There are no robust `pass` (or any other) tools can help to print passwd for himalaya. (The `echo` command in `cmd.exe` prints newline.)

I would suggest adding an alternative way of putting the string as the passwd directly. If someone cares about secret leak they can still use the `cmd` way.

soywod:
@chawyehsu the `echo` command also add a new line, and it is trim out by the tool:
https://github.com/soywod/himalaya/blob/d5a79ed6dba2f23f2ded95ab7d02e4e4227c994f/src/config/model.rs#L50

If you still have problems with the new line, then you can add the flag `-n` which removes the new line (it is the same on linux).

So I do not see the need to add a new entry in the config.

chawyehsu:
**Did you test it on a Windows machine?** I assume that you were not: 
```
❯ himalaya.exe
Cannot login to IMAP server
 ↳ Invalid character in input: '\r'
```

[...]

PR #111, opened 2021-04-19
fix: fix config loading compatibility on Windows

chawyehsu:
- Fix https://github.com/soywod/himalaya/issues/72#issuecomment-822101907
- Use `USERPROFILE` as alternative HOME on Windows

### DIFF
--- a/src/config/model.rs
+++ b/src/config/model.rs
@@ -119,10 +119,14 @@ impl Config {
     }
 
     fn path_from_xdg_alt() -> Result<PathBuf> {
-        let path = env::var("HOME")
-            .or(env::var("USERPROFILE"))
-            .chain_err(|| "Cannot find `HOME` env var")?;
-        let mut path = PathBuf::from(path);
+        let home_var = if cfg!(target_family = "windows") {
+            "USERPROFILE"
+        } else {
+            "HOME"
+        };
+        let mut path: PathBuf = env::var(home_var)
+            .chain_err(|| format!("Cannot find `{}` env var", home_var))?
+            .into();
         path.push(".config");
         path.push("himalaya");
         path.push("config.toml");
@@ -131,10 +135,14 @@ impl Config {
     }
 
     fn path_from_home() -> Result<PathBuf> {
-        let path = env::var("HOME")
-            .or(env::var("USERPROFILE"))
-            .chain_err(|| "Cannot find `HOME` env var")?;
-        let mut path = PathBuf::from(path);
+        let home_var = if cfg!(target_family = "windows") {
+            "USERPROFILE"
+        } else {
+            "HOME"
+        };
+        let mut path: PathBuf = env::var(home_var)
+            .chain_err(|| format!("Cannot find `{}` env var", home_var))?
+            .into();
         path.push(".himalayarc");
 
         Ok(path)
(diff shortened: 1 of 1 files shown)


src/config/model.rs
```diff
@@ -47,7 +47,7 @@ impl Account {
 
     pub fn imap_passwd(&self) -> Result<String> {
         let passwd = run_cmd(&self.imap_passwd_cmd).chain_err(|| "Cannot run IMAP passwd cmd")?;
-        let passwd = passwd.trim_end_matches("\n").to_owned();
+        let passwd = passwd.trim_end().to_owned();
```
soywod, line 50: With `trim_end` you also remove spaces, which are *in theory* valid chars for a password. I admit that *in practice* nobody uses this char in his password, but still. What do you think about:

```rust
let passwd = passwd
    .trim_end_matches(|c| c == '\r' || c == '\n')
    .to_owned();
```

src/config/model.rs
```diff
@@ -68,7 +68,7 @@ impl Account {
 
     pub fn smtp_creds(&self) -> Result<SmtpCredentials> {
         let passwd = run_cmd(&self.smtp_passwd_cmd).chain_err(|| "Cannot run SMTP passwd cmd")?;
-        let passwd = passwd.trim_end_matches("\n").to_owned();
+        let passwd = passwd.trim_end().to_owned();
```
soywod, line 71: With `trim_end` you also remove spaces, which are *in theory* valid chars for a password. I admit that *in practice* nobody uses this char in his password, but still. What do you think about:

```rust
let passwd = passwd
    .trim_end_matches(|c| c == '\r' || c == '\n')
    .to_owned();
```

src/config/model.rs
```diff
@@ -115,7 +115,9 @@ impl Config {
     }
 
     fn path_from_xdg_alt() -> Result<PathBuf> {
-        let path = env::var("HOME").chain_err(|| "Cannot find `HOME` env var")?;
+        let path = env::var("HOME")
+            .or(env::var("USERPROFILE"))
+            .chain_err(|| "Cannot find `HOME` env var")?;
```
soywod, line 120: `USERPROFILE` is only used by Windows, right? So I will go with sth more concise, like:

```rust
let home_var = if cfg!(target_family = "windows") {
    "USERPROFILE"
} else {
    "HOME"
};
let home_path: PathBuf = env::var(home_var)
    .chain_err(|| format!("Cannot find `{}` env var", home_var))?
    .into();
```

See https://doc.rust-lang.org/reference/conditional-compilation.html#target_family.

src/config/model.rs
```diff
@@ -125,7 +127,9 @@ impl Config {
     }
 
     fn path_from_home() -> Result<PathBuf> {
-        let path = env::var("HOME").chain_err(|| "Cannot find `HOME` env var")?;
+        let path = env::var("HOME")
+            .or(env::var("USERPROFILE"))
+            .chain_err(|| "Cannot find `HOME` env var")?;
```
soywod, line 132: `USERPROFILE` is only used by Windows, right? So I will go with sth more concise, like:

```rust
let home_var = if cfg!(target_family = "windows") {
    "USERPROFILE"
} else {
    "HOME"
};
let home_path: PathBuf = env::var(home_var)
    .chain_err(|| format!("Cannot find `{}` env var", home_var))?
    .into();
```

See https://doc.rust-lang.org/reference/conditional-compilation.html#target_family.

soywod (changes requested):
Thanks for your contribution! I had some tiny changes, let me know what you think about them.

src/config/model.rs
chawyehsu: Make sense, I will update the patch.

soywod (approved):

soywod:
Thanks :tada:

PR #114, opened 2021-04-19
Added signature separator option plus some refactoring

### LINKED ISSUE #113: Add option for separator between content and signature

AckslD:
closes #113 

I started adding an option to change the default '--' to separate the content and signature of a message. While doing so I also merge some common code for the `build_xxx` functions.

### DIFF
--- a/src/config/model.rs
+++ b/src/config/model.rs
@@ -19,7 +19,6 @@ pub struct Account {
     pub name: Option<String>,
     pub downloads_dir: Option<PathBuf>,
     pub signature: Option<String>,
-    pub signature_separator: Option<String>,
     pub default_page_size: Option<usize>,
 
     // Specific
@@ -102,7 +101,6 @@ pub struct Config {
     pub downloads_dir: Option<PathBuf>,
     pub notify_cmd: Option<String>,
     pub signature: Option<String>,
-    pub signature_separator: Option<String>,
     pub default_page_size: Option<usize>,
 
     #[serde(flatten)]
@@ -214,14 +212,6 @@ impl Config {
             .signature
             .as_ref()
             .or_else(|| self.signature.as_ref())
-            .map(|sig| sig.to_owned())
-    }
-
-    pub fn signature_separator(&self, account: &Account) -> Option<String> {
-        account
-            .signature_separator
-            .as_ref()
-            .or_else(|| self.signature_separator.as_ref())
             .map(|sig| sig.to_owned())
     }
 
--- a/src/config/model.rs
+++ b/src/config/model.rs
@@ -19,6 +19,7 @@ pub struct Account {
     pub name: Option<String>,
     pub downloads_dir: Option<PathBuf>,
     pub signature: Option<String>,
+    pub signature_separator: Option<String>,
     pub default_page_size: Option<usize>,
 
     // Specific
@@ -101,6 +102,7 @@ pub struct Config {
     pub downloads_dir: Option<PathBuf>,
     pub notify_cmd: Option<String>,
     pub signature: Option<String>,
+    pub signature_separator: Option<String>,
     pub default_page_size: Option<usize>,
 
     #[serde(flatten)]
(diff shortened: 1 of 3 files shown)


src/config/model.rs
```diff
@@ -215,6 +217,14 @@ impl Config {
             .map(|sig| sig.to_owned())
     }
 
+    pub fn signature_separator(&self, account: &Account) -> Option<String> {
```
soywod, line 220: In fact a signature separator fn does not make sense (it will never be used alone). It makes more sense *I think* to move this part inside the `signature` fn. What do you think?

soywod (changes requested):
I like the way you refactored, it looks cleaner to me!

src/msg/model.rs
```diff
@@ -499,86 +441,123 @@ impl<'m> Msg<'m> {
...
 
+    fn build_tpl(config: &Config, account: &Account, msg_spec: MsgSpec) -> Result<Tpl> {
+        let mut tpl = vec![];
+        Msg::add_content_headers(&mut tpl);
+        Msg::add_from_header(&mut tpl, Some(config.address(account)));
+        Msg::add_in_reply_to_header(&mut tpl, msg_spec.in_reply_to);
+        Msg::add_cc_header(&mut tpl, msg_spec.cc);
+        Msg::add_to_header(&mut tpl, msg_spec.to);
+        Msg::add_subject_header(&mut tpl, msg_spec.subject);
+        tpl.push(String::new()); // Separator between headers and body
```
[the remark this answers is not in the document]
AckslD: Sounds good! I moved it inside the if-statement in `add_content` such that it is only added if there actually is some additional content. Do you agree with this? (previously for new emails there were two empty lines I think).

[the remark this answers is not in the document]
soywod: Sounds great!

src/config/model.rs
AckslD: Yeah maybe. The only thing I'm wondering if handling defaults here is desired? Cause I think if it's moved into `signature` then the two `Optional<String>`s need to be merged somehow, either by replacing them by defaults if `None` or returning `(Option<String>, Option<String>)`. What do you think?

soywod: Indeed, I did not think about this. After all, maybe the simplest way is to include the separator *inside* the signature. The user needs to add the separator *inside* his signature (if he wants one). It just needs to be better documented. What do you think?

AckslD: Ah yes, this is probably the way to go :) I just tried that one can indeed do multiline strings in the config 👍

AckslD: Removed the option again. I made the example config in the README use a multiline signature, do you think that's useful?

AckslD: I added an explanation in the wiki (didn't know anyone could edit it :))

soywod: Oh great! Thank you :smiley:

soywod:
@AckslD In fact, the double dash seems to be [a standard](https://en.wikipedia.org/wiki/Signature_block#Standard_delimiter). In the `.signature` file, the dashes should not be included because they are added automatically. I tried with emacs and thunderbird and both have the same behaviour. I think we should reconsider this PR to align the tool with standards.

I propose sth close to what you originally proposed: we introduce a new boolean `signature-delimiter`. If true (the default) then we show the delimiter before the signature. If not then only the signature is shown. Those who want a custom delimiter can disable the `signature-delimiter` and include it inside their signature. What do you think?

AckslD:
@soywod I see, yeah that make sense to me :+1:

soywod:
@AckslD I pushed the changes (a6b30b746c3a2d77eb2e48bd7d7cb003cf88cd8a). I introduced `signature-delimiter` as on optional string instead of a boolean, I thought it would be more UX. In fact it really looks like your initial PR... sorry for all those changes.

---
It's been a while I did not see you here, I hope your are doing well!

AckslD:
Nice @soywod! :)

---

I started a new job so been quite busy with that :) Hope you're doing well also!

PR #116, opened 2021-04-20
handle completions subcommand first

### LINKED ISSUE #115: Generate shell completions without mail configure
Hi there,

Himalaya is packing into nix package manage,  and I got the issue on generating the completions without mail configure.

https://github.com/NixOS/nixpkgs/pull/119852

So I change the subcommand handle process in the main loop.

If there is any concern please kindly tell me.

Thanks.

yanganto:
fix #115

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -67,7 +67,7 @@ fn run() -> Result<()> {
     debug!("[main] log level: {}", log_level);
     debug!("[main] custom config path: {:?}", custom_config);
 
-    if let Ok(_matched) = comp_matches(build_app(), &matches) {
+    if comp_matches(build_app(), &matches)? {
         return Ok(());
     }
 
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -67,6 +67,10 @@ fn run() -> Result<()> {
     debug!("[main] log level: {}", log_level);
     debug!("[main] custom config path: {:?}", custom_config);
 
+    if let Ok(_matched) = comp_matches(build_app(), &matches) {
```
soywod, line 70: By doing so you ignore all potential errors. You need to keep the `?` operator so that error are handled properly. Sth like:

```suggestion
    if comp_matches(build_app(), &matches)? {
```

soywod (changes requested):

soywod:
Thank you :tada:

PR #128, opened 2021-04-24
add content-transfer-encoding support

### LINKED ISSUE #127: Content-Transfer-Encoding is not correct
The mail template use 8bit for encoding
https://github.com/soywod/himalaya/blob/master/src/msg/model.rs#L492
```rust
        tpl.push("Content-Transfer-Encoding: 8bit".to_string());
```

But the real encoding for the mail is base64
https://github.com/soywod/himalaya/blob/master/src/msg/model.rs#L293
```rust
        let text_part = SinglePart::builder()
            .header(ContentType("text/plain; charset=utf-8".parse().unwrap()))
            .header(ContentTransferEncoding::Base64)
            .body(parsed.get_body_raw()?);
```

yanganto:
fix #127

### DIFF
--- a/src/msg/model.rs
+++ b/src/msg/model.rs
@@ -1,5 +1,6 @@
 use error_chain::error_chain;
 use lettre;
+use log::warn;
 use mailparse::{self, MailHeaderMap};
 use rfc2047_decoder;
 use serde::{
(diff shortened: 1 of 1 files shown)


soywod:
Thanks a lot :tada:

PR #161, opened 2021-06-03
Add .desktop file

### LINKED ISSUE #160: Add `share/application/himalaya.desktop`
This file would be really nice to have specifically so users can leverage XDG and set `himalaya` as a default mail app

```
[Default Applications]
x-scheme-handler/mailto=himalaya.desktop;
```

as well as getting access to be chosen as the default for Firefox, `xdg-email`, etc.

- - -

I think literally this could actually be a static file @ `$PROJECT_ROOT/share/application/himalaya.desktop` and package managers can just copy it. If you don't have the bandwidth @soywod, I could probably do this (just tap "Assign").

toastal:
fixes #160

It seems other projects actually just put this in an `assets/` directory. This would probably be the place say a logo would go. 

The flake could maybe be better overriding the `Exec` with the location in the store, but I think this is good enough and it gives package maintainers a good start.

The `Terminal=true` value makes it be launched in the user's configured terminal

### DIFF
--- a/assets/himalaya.desktop
+++ b/assets/himalaya.desktop
@@ -1,0 +1,18 @@
+[Desktop Entry]
+Type=Application
+Name=himalaya
+DesktopName=Himalaya
+GenericName=Mail Reader
+Comment=CLI email client written in Rust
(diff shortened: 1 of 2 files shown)


soywod:
Thanks a lot :tada:

Issue #156, opened 2021-05-23
Cache sessions in a daemon


hardselius:
First of all, this is a fantastic project! I've been toying around with it for a few days.

I use `pass` to store my passwords and I immediately noticed that it introduces a significant delay when used as a password command. It would be great to have the option to cache credentials between executions for when calculating credentials is slow.

soywod:
It is really strange because I also use `pass` as password command and I do not have any delay. Could it be related to your configuration (pass, gpg)?

> It would be great to have the option to cache credentials between executions for when calculating credentials is slow

In a more generic way, I was thinking about adding an option to spawn a daemon that connects to your IMAP server and to wait for commands from the CLI. This way we keep all the login process in memory. But it is clearly not a priority, I will keep this issue for after the `v1.0.0`. Thanks for your feedback!

hardselius:
Peculiar. I'm using a Yubikey, so that might be what's causing the delay. I will have to investigate. 

The daemon sound like a nice addition though.

soywod:
Now that himalaya supports Maildir and Notmuch, I'm not so sure about such daemon. Plus the amount of work is giant compare to the benefits. The only place I still see the feature is in the TUI, but even there we don't need a *daemon*, we can just keep in memory the session while the process runs.

PR #155, opened 2021-05-21
TUI

TornaxO7:
You can view in this pull request the "process" of the implementation of the TUI. Pls consider that I'm completely new in rust so I might be pretty slow, but I'll do my best.

# Status
Created the `src/tui` directory with

```bash
cargo new --lib tui
```

# Information for maintainer (@soywod )
I'm creating the pull request now, so you can review any time the latest commits whenever you want, so you can give me immediately a feedback if something doesn't suit you. Is that fine for you?

PR #170, opened 2021-06-20
Cli refactor

TornaxO7:
I need to change some stuff in the current API because I need to be able to cache the data to improve the speed in the TUI. But before that I changed the CLI-Stuff a little bit:

Moved all cli.rs files into the new directory src/cli. In my opinion this makes it more clear where the functions like cli::flag::subcmds() are coming from
Added a documentation to all files in src/cli
This is just a suggestion and not really neccessary (except for the documentation :D). I'm open for discussions about this!

Please do cargo doc --open and look what I've written there to understand "faster" what I've changed.

(Hint: I renamed the branch to `cli-refactor`, that's why there's a "second" (closed) pull request)


soywod:
> I need to change some stuff in the current API because I need to be able to cache the data to improve the speed in the TUI.

I'm sorry in advance if we already discuss about it, I may have forgotten. My moving is in 2 weeks and my mind is(will) be away for a while :smile:. I'm not sure the cache issue should be included in your PR. Because it is not related to the TUI (we have the same problem with the CLI). I think this should be put aside in a dedicated issue (in fact we can use #156).

> Moved all cli.rs files into the new directory src/cli. In my opinion this makes it more clear where the functions like cli::flag::subcmds() are coming from

When I started Himalaya, I asked myself this question: should I organize stuff by domain (business logic like imap, mbox, msg) or not (technical logic like cli, tui). And I took the decision to organize by domain (I will try to find back the article that made my decision. The core idea is to reduce distance between dependencies, everything belonging to the same logic should be as close as possible to each other to keep things simple and understandable). By refactoring so, you break a bit this decision. I would expect a `tui.rs` file by business module.

> Added a documentation to all files in src/cli
> This is just a suggestion and not really neccessary (except for the documentation :D). I'm open for discussions about this!

The documentation is a very good initiative, thank you!

TornaxO7:
> My moving is in 2 weeks and my mind is(will) be away for a while smile.

No problem

> I'm not sure the cache issue should be included in your PR. Because it is not related to the TUI (we have the same problem with the CLI). I think this should be put aside in a dedicated issue (in fact we can use #156).

Yes, that's why I want to create a second pull request (currently on it).

> By refactoring so, you break a bit this decision. I would expect a tui.rs file by business module.

Wait, I think you missunderstood something here (or I'm missunderstanding you). This pull request has nothing to do with the TUI. It was just a starting suggestion to group all cli files into one directory.

What I mean with "caching" is actually making each struct independent of the imap connection. Because if we take a look at `Msg` for example, than we can see that it can be only used as long the imap-connection exists (due to the `flags` attribute). This makes it "impossible" for me to store the mails in the memory, so I'd just like to "refactor" them by implementing a way, to "extract" the data from the imap connection and store them "independently" in these structs. Or should I rather wait for the daemon implementation first?

TornaxO7:
Should I close this pull request then?

By the way, the doc won't work anymore, because it focuses on the "new" `src/cli` directory.

soywod:
> Wait, I think you missunderstood something here (or I'm missunderstanding you). This pull request has nothing to do with the TUI. It was just a starting suggestion to group all cli files into one directory.

The `tui.rs` was just an example. The fact to put stuff in a `cli` folder breaks the "splitting by domain" rule. Instead I see more a `cli.rs` in a `msg` folder, or a `tui.rs` in a `mbox` folder.

About caching, see my reply on #155. I don't think it is sth to consider _right now_. I agree that having a TUI without caching does not make sense unless you use an IMAP synchronizer. I was planning to fix this caching issue _before_ implementing the TUI, but you proposed yourself for the TUI (and I will never thank you enough for this initiative :wink:).

TornaxO7:
> Should I close this pull request then?

So I think you also mean to say: Yes ;)

PR #171, opened 2021-06-23
Add Message-Id header to msg

fancycade:
Hi! 

I ran into an issue where my emails were being rejected because of a missing Message-Id header.

Since lettre has support for this it was trivial to fix.

Here is a [link](https://github.com/lettre/lettre/blob/0439bab87481dc7ce30c8a79492f14b729df6608/src/message/mod.rs#L357) to lettre's implementation/docs.

~~The default option with None works fine, but it will use the local hostname as the suffix.
Perhaps a better solution would be to use the From header's domain.~~

Message-Id is generated from a uuid and the from address domain.

Happy to make changes as necessary!

### DIFF
--- a/src/msg/model.rs
+++ b/src/msg/model.rs
@@ -291,8 +291,11 @@ impl<'m> Msg<'m> {
 
             match h.get_key().to_lowercase().as_str() {
                 "in-reply-to" => msg.in_reply_to(value.parse().unwrap()),
-                "from" => match value.parse() {
-                    Ok(addr) => msg.from(addr),
+                "from" => match value.parse::<lettre::message::Mailbox>() {
+                    Ok(addr) => {
+                        let msg_id = format!("{}@{}", Uuid::new_v4().to_string(), addr.email.domain());
+                        msg.from(addr).message_id(Some(msg_id))
+                    }
                     Err(_) => msg,
                 },
                 "to" => value


soywod:
Thanks :tada:!

TornaxO7:
Wait! Shouldn't the message-id be set by the host? 

> The "Message-ID:" field provides a unique message identifier that
   refers to a particular version of a particular message.  The
   uniqueness of the message identifier is guaranteed by the host that
   generates it (see below).

Quote from [here](https://datatracker.ietf.org/doc/html/rfc5322#section-3.6.4).

fancycade:
@TornaxO7 My host (migadu) didn't seem to set it, so I don't think it can be guaranteed that the host will do it.

TornaxO7:
Well, then I think that this is a bug by the host. As the RFC points out, it's rather a job for the host not for the client to set the message id.

TornaxO7:
Ok, so re-read the RFC again. If I understood it right, we have two message-id headers:
- `message-id` and
- `msg-id`

`message-id`:
This is set from the host, but this can be also set of the `lettre` crate.

`msg-id`:
Can be set from the client (and from the host?) and is needed for the host.

So I think, that the problem here is, that `lettre` uses the wrong `message-id` field to set the id? Am I right? Can anyone comfirm my theory?

soywod:
@paolobarbolini?

fancycade:
From my research it is best practice for the client to generate the message id

Here it is being set in the alps client:

https://git.sr.ht/~emersion/alps/tree/master/item/plugins/base/smtp.go#L142

I would imagine I can find this in more email clients.

Also this stack overflow answer:
https://stackoverflow.com/questions/14483861/whats-the-issue-with-message-id-in-email-sent-by-php

It seems that some hosts will generate one if it isn't already set.

TornaxO7:
Now I'm even more irritated. The stack overflow question uses the `Message-ID` header but the value is the same as for the `msg-id` header???

TornaxO7:
According to the [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322#section-3.6.4) `msg-id` should have this "style" with these `<`,`>` brackets and the `@` between the random id on the left and the hostname on the right while the `Message-ID` header should have an unique identifier.

TornaxO7:
Huh, so I read one of my mails from gitter raw (through `himalaya read --raw <UID>`) and I'm getting this:

```
Message-ID: <01000179146aacc6-46e43009-0c3a-45ea-885e-447cf1bcf694-000000@email.amazonses.com>
```

Looks like that `msg-id` is supposed to be `Message-ID`??

TornaxO7:
Oh hold on, I'm stupid

image

`msg-id` is the *value* for `Message-ID`..... daaaaaamn, sorry! My mistake here! I have to fix that in my pull request as well!

PR #182, opened 2021-08-03
Fix Vim plugin + update Flake references

toastal:
The Vim plugin was failing to build. I opened an upstream [merge request](https://github.com/NixOS/nixpkgs/pull/132517) in `nixpkgs` as well.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -41,26 +41,24 @@
             generatedCargoNix;
 
           # Create the cargo2nix project
-          project = pkgs.callPackage
-            (generatedCargoNix {
-              inherit name;
-              src = ./.;
-            })
-            {
-              # Individual crate overrides go here
-              # Example: https://github.com/balsoft/simple-osd-daemons/blob/6f85144934c0c1382c7a4d3a2bbb80106776e270/flake.nix#L28-L50
-              defaultCrateOverrides = pkgs.defaultCrateOverrides // {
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -70,8 +74,12 @@
               inherit (packages.${name}) version;
               name = "${name}-vim";
               src = self;
-              configurePhase = "cd vim/";
               buildInputs = [ packages.${name} ];
+              patchPhase = ''
```
soywod, line 78: Could you explain those lines? I mean, why they are necessary?

soywod (changes requested):

flake.nix
toastal: I understand now what happened so I did I different fix (the wrapper for building vim plugins upstream changed).

However, `''\n''` is lke starting a heredoc; it lets you write write a longer inline `bash` script in this block which if you have literally more than one line, is recommended.

toastal:
I was pointed in the direction of the upstream Vim plugin building wrapper now by default disabling the configure phase with `dontConfigure = true;` being the default. This is really weird to me because there already was a check and the default is `configurePhase ? ""`.

toastal:
Now, why Cachix is failing in some JS script is :shrug: to me. I don't know or want to know GitHub Actions. My upstream Vim plugin fix was merged and this builds for me locally.

soywod:
Thanks :tada:

PR #194, opened 2021-08-29
fix read/reply in vim plugin

mtoohey31:
Hello! I was encountering issues with reading and repling in the vim plugin, which a I believe were a result of a recent refactor removing the `"raw"` property in the `json` output format of commands for getting templates. The fix is to use the `plain` output format instead, this required some modifications to the `himalaya#shared#cli#call` function though. In order to avoid having to change all of the calls to this function, I added the format to be used as an optional argument that defaults to `"json"`, however, if you'd like I can switch it to a required argument and change all other calls to specify `"json"` as the required format. Or, if there's a different way you'd like this fixed, let me know and I can get that done too.


soywod:
A big PR is coming soon about the `Msg` refactor, maybe it will fix partially this issue. I will wait for the merge and come back to you. Thank you for your help!

TornaxO7:
Hi! The Msg refactoring PR has been merged. Could you please try it again and look, if the issue still exist?

mtoohey31:
Sure, I'll do my best to get to it as soon as I can, school just started though so it might be a few days. Also, thanks for your work on #155, I'm really looking forward to it!

mtoohey31:
Sorry, just getting to this now, could you point me in the direction of the commit I should use to test this? I tried 0e68801, but I'm running into a `Couldn't get the data from the 'To:' field.` error just when running himalaya on its own, and on the latest commit, get `Error: cannot extract recipients from envelope`.

soywod:
It seems that the vim plugin is not working as expected since last merges. I finish some cleaning and I take care of it!

mtoohey31:
I don't mind working on the vim plugin, if you'd like me to, like I mentioned though, those errors occurred when running himalaya _on its own_, normally, from the command line, and debugging that is a bit out of my depth as I'm not that experienced with rust.

soywod:
I come back to you once the cleaning done, it will be easier to move forward. I should (hope) to finish within few days!

PR #195, opened 2021-08-29
disable cursorline for read/write in vim plugin

mtoohey31:
Hello! This PR is more of an opinion thing than a bug fix, but the vim plugin enables the `cursorline` option for the `himalaya-message-list` filetype, as seen below, which makes sense:

image_2021-08-29-16-23-49

However, it also does the same thing for both reading and writing messages, which is something that I believe should be left up to the user, as having cursorline enabled in these buffers is really more of an editor preference thing than it is a UI as in the case of the `himalaya-message-list` filetype. Here are the corresponding screenshots:

image_2021-08-29-16-24-49
image_2021-08-29-16-26-41

If this PR is merged and someone prefers `cursorline` in their read and write buffers, they can easily set that preference themself by placing the following line in their vim/neovim configuration:

```vim
autocmd FileType mail setlocal cursorline
```

### DIFF
--- a/vim/ftplugin/himalaya-msg-read.vim
+++ b/vim/ftplugin/himalaya-msg-read.vim
@@ -1,6 +1,5 @@
 setlocal bufhidden=wipe
 setlocal buftype=nofile
-setlocal cursorline
 setlocal filetype=mail
 setlocal foldexpr=himalaya#shared#thread#fold(v:lnum)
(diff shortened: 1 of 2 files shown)


soywod:
I totally agree with you. Thanks for your contribution!

mtoohey31:
:+1: Happy to help!

Issue #172, opened 2021-06-23
Could not append message to `Sent`


TornaxO7:
So I'm testing some stuff currently and when I tried `himalaya write`, this happened:
```
Could not append message to `Sent`
 ↳ No Response: Folder doesn't exist. (Failure)
```
But: If I look into my gmail account, I cann see my mail under the `Send` mailbox. So doesn't that mean that this error message is wrong?

This is my `~/.config/himalaya/config.toml`:
```toml
name = "Sambalakuja"
downloads-dir = "~/"
default-page-size = 50
signature = "I use Arch btw"
watch-cmds = ["mbsync -a"]

[gmail]
default = true
email = "Sambalakuja@gmail.com"

imap-host = "imap.gmail.com"
imap-port = 993
imap-login = "sambalakuja@gmail.com"
imap-passwd-cmd = "pass show sambalakuja@gmail.com"

smtp-host = "smtp.gmail.com"
smtp-port = 465
smtp-login = "sambalakuja@gmail.com"
smtp-passwd-cmd = "pass show sambalakuja@gmail.com"
```

soywod:
It may be because of the `~` in the `downloads-dir`. As described in the wiki, you need to put the full abs path. See also #102.

TornaxO7:
I changed it to this:
```toml
name = "Sambalakuja"
downloads-dir = "/home/tornax/Downloads"
default-page-size = 50
signature = "I use Arch btw"
watch-cmds = ["mbsync -a"]

[gmail]
default = true
email = "Sambalakuja@gmail.com"

imap-host = "imap.gmail.com"
imap-port = 993
imap-login = "sambalakuja@gmail.com"
imap-passwd-cmd = "pass show sambalakuja@gmail.com"

smtp-host = "smtp.gmail.com"
smtp-port = 465
smtp-login = "sambalakuja@gmail.com"
smtp-passwd-cmd = "pass show sambalakuja@gmail.com"
```
But the error still exists.

soywod:
So I guess it comes from the `Sent` folder from the IMAP server. When you run the command `himalaya mbox`, do you see the `Sent` mailbox?

TornaxO7:
That's what I get if I run `himalaya mailboxes`:

```rust

[1m[4m[37mDELIM [0m[38;5;8m│[0m[1m[4m[37mNAME              [0m[38;5;8m│[0m[1m[4m[37mATTRIBUTES                 [0m
[...]

PR #226, opened 2021-10-14
Adding new enum to detect the type of a mailbox

TornaxO7:
This should fix https://github.com/soywod/himalaya/issues/172

### DIFF
--- a/src/domain/msg/msg_entity.rs
+++ b/src/domain/msg/msg_entity.rs
@@ -17,7 +17,7 @@ use crate::{
     config::Account,
     domain::{
         imap::ImapServiceInterface,
-        mbox::{Mbox, Mboxes, MboxType},
+        mbox::{Mboxes, MboxType},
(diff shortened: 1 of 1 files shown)


soywod:
I'm afraid you started to develop this from `master`, so you are not aware about new stuff in `development`. I changed the target branch and you seem to have conflicts. I refactored well the `Mbox` (with comments + tests), maybe this refactor will help you in what you wanted to achieve there? It sounds great to me what you did anyway, thanks to take this point it's a tough one

soywod:
Oh wait, this means we will have one extra call every time :confused:. The user does not have to configure anything at the cost of the response delay. But what if a user defines a mailbox named "MySent" with attributes "Custom(MySent)" as his sending folder? This case would not work. Maybe we should let users configure their mailboxes name. We could set up defaults (as it works right now on master), I guess it will match 80% of users.

PR #248, opened 2021-10-29
add always a newline for the outputs.

gkmngrgn:
Hi @soywod,

First of all, thank you for this -accessibility friendly- project. I'm using it actively and it works awesome even with some mistakes.

Here you can see a screenshot of why I did open this open pull request:
image

ZSH always adds a strange character `%` to the end of the output messages. I also checked if my change broke anything in bash; it seems everything is fine.

What do you think?

Regards.


src/output/print.rs
```diff
@@ -9,7 +9,7 @@ pub trait Print {
 
 impl Print for &str {
     fn print(&self, writter: &mut dyn WriteColor) -> Result<()> {
-        write!(writter, "{}", self).with_context(|| {
+        write!(writter, "{}\n", self).with_context(|| {
```
soywod, line 12: There is a macro `writeln!` that is adding a new line for you, I would advise to use it instead.

soywod (changes requested):
I switched from bash to zsh recently and I did not notice this character. But a new line costs nothing, let's merge after your changes. Thanks for your feedback!

src/output/print.rs
gkmngrgn: The new PR is ready.
https://github.com/soywod/himalaya/pull/251

PR #251, opened 2021-10-31
add always a newline for the outputs.

gkmngrgn:
Hi @soywod, this is my second PR because the first one (#248) doesn't allow me to change the source branch from `master` to `development`. In this  PR:

- the target branch is `soywod/development`.
- I use `writeln!` instead of `{}\n`.
- I also tested with zero ZSH configuration. I still see % in the same line, but this PR **will fix** this problem.

image

Thanks. I'll close #248.


soywod:
Thank you :tada:

PR #255, opened 2021-11-02
clean: fix clippy lints

Ethiraric:
This commit should have no impact on the behavior of the program.

I couldn't find any documentation on what the testing procedure is. Both `cargo test` and `cargo test --release` are happy.
Feel free to point me towards more in-depth tests.

### DIFF
--- a/email-manager/src/domain/msg/msg_handler.rs
+++ b/email-manager/src/domain/msg/msg_handler.rs
@@ -21,6 +21,7 @@ use crate::{
         mbox::Mbox,
         msg::{Flags, Msg, Part, TextPlainPart},
         smtp::SmtpServiceInterface,
+        Parts,
     },
     output::{PrintTableOpts, PrinterService},
 };
@@ -32,7 +33,7 @@ pub fn attachments<'a, Printer: PrinterService, ImapService: ImapServiceInterfac
     printer: &mut Printer,
     imap: &mut ImapService,
 ) -> Result<()> {
-    let attachments = imap.find_msg(&seq)?.attachments();
+    let attachments = imap.find_msg(seq)?.attachments();
     let attachments_len = attachments.len();
     debug!(
         r#"{} attachment(s) found for message "{}""#,
@@ -60,7 +61,7 @@ pub fn copy<'a, Printer: PrinterService, ImapService: ImapServiceInterface<'a>>(
     imap: &mut ImapService,
 ) -> Result<()> {
     let mbox = Mbox::new(mbox);
-    let msg = imap.find_raw_msg(&seq)?;
+    let msg = imap.find_raw_msg(seq)?;
     let flags = Flags::try_from(vec![Flag::Seen])?;
     imap.append_raw_msg_with_flags(&mbox, &msg, flags)?;
     printer.print(format!(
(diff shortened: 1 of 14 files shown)


email-manager/src/main.rs
```diff
@@ -60,11 +58,8 @@ fn main() -> Result<()> {
 
     // Check completion command BEFORE entities and services initialization.
     // Related issue: https://github.com/soywod/himalaya/issues/115.
-    match compl_arg::matches(&m)? {
-        Some(compl_arg::Command::Generate(shell)) => {
-            return compl_handler::generate(create_app(), shell);
-        }
-        _ => (),
+    if let Some(compl_arg::Command::Generate(shell)) = compl_arg::matches(&m)? {
```
soywod, line 61: I would prefer to keep the original version in case we add more commands in the future (plus to match the other matches style).

email-manager/src/main.rs
```diff
@@ -87,11 +82,8 @@ fn main() -> Result<()> {
     }
 
     // Check mailbox commands.
-    match mbox_arg::matches(&m)? {
-        Some(mbox_arg::Cmd::List(max_width)) => {
-            return mbox_handler::list(max_width, &mut printer, &mut imap);
-        }
-        _ => (),
+    if let Some(mbox_arg::Cmd::List(max_width)) = mbox_arg::matches(&m)? {
```
soywod, line 85: I would prefer to keep the original version in case we add more commands in the future (plus to match the other matches style).

email-manager/src/domain/msg/msg_handler.rs
```diff
@@ -161,16 +162,18 @@ pub fn mailto<
...
+    let msg = Msg {
+        from: Some(vec![account.address().parse()?]),
+        to: if to.is_empty() { None } else { Some(to) },
+        cc: if cc.is_empty() { None } else { Some(cc) },
+        bcc: if bcc.is_empty() { None } else { Some(bcc) },
+        subject: subject.into(),
+        parts: Parts(vec![Part::TextPlain(TextPlainPart {
+            content: body.into(),
+        })]),
+        ..Default::default()
```
soywod, line 174: Is it a common practice in Rust to use the `Default::defaut()` instead of `Msg::default()`?

soywod (changes requested):
Thanks a **lot** for all these details, the code looks fresher :sparkles:

soywod:
> I couldn't find any documentation on what the testing procedure is.

It is part of sth I would like to set up: some kind of "How to contribute". I created an issue not to forget this point https://github.com/soywod/himalaya/issues/256.

email-manager/src/main.rs
Ethiraric: Added a `#[allow(clippy::single_match)]` above the main function.

email-manager/src/domain/msg/msg_handler.rs
Ethiraric: I let clippy's fix-it correct it for me and it automatically replaced `Msg` with `Default`. In all honesty, I wasn't aware of that syntax (using `..` in a struct instantiation).

I have no idea what the best practice on that is. It works with `Msg::default()` so I changed it from `Default` to `Msg`.

soywod:
Thanks :tada:

soywod:
A contributing guide has been added on `development`, it will be available starting from the next release!

PR #277, opened 2022-01-17
Use regex to find focused message id

kreemer:
Dears maintainers,

This PR changes the way, how the message id in the line is parsed. I was getting issues when using the "split()" variant because it could not split by the pipe character (i'm dunno why).

With this change, the message id will be parsed using a regex.


soywod:
In fact, the issue has already been fixed by someone else https://github.com/soywod/himalaya/pull/247. The fix is on the `development` branch, it will be updated with the next release.

PR #293, opened 2022-02-09
Improve attachment output

### LINKED ISSUE #281: Improve attachments command
`himalaya attachments n` currently downloads attachments and displays the number of files downloaded.
I'd like to see the file names in output, that would help to find them in big folders, eventually with a prompt to change the filename if it already exists in the downloads-dir.

palb91:
Fix #281 

This is a small PR to display the filename of the attachments downloaded. 

However, I have a warning when running `cargo check`:

```rust
warning: unused `Result` that must be used
  --> src/domain/msg/msg_handler.rs:46:9
   |
46 |         printer.print(format!("downloading {}…", attachment.filename));
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_must_use)]` on by default
   = note: this `Result` may be an `Err` variant, which should be handled

warning: `himalaya` (bin "himalaya") generated 1 warning
```

I understand that Rust uses some structures that auto-handle Success/Error, but I don't get why there is no warning for the same function on line 51. I have found this [SO thread](https://stackoverflow.com/questions/63437768/unused-stdresultresult-that-must-be-used-when-using-writeln) but it feels really redundant with the debug line.

This is my first contact with Rust, but it shouldn't be too hard to fix ^^

### DIFF
--- a/src/domain/msg/msg_handler.rs
+++ b/src/domain/msg/msg_handler.rs
@@ -43,6 +43,7 @@ pub fn attachments<'a, Printer: PrinterService, ImapService: ImapServiceInterfac
     for attachment in attachments {
         let filepath = account.downloads_dir.join(&attachment.filename);
         debug!("downloading {}…", attachment.filename);
+        printer.print(format!("downloading {}…", attachment.filename));
         fs::write(&filepath, &attachment.content)
             .context(format!("cannot download attachment {:?}", filepath))?;
     }


soywod:
First of all, thank you for contributing to himalaya :pray:

You get this warning because `printer.print` returns a `Result<…>` and needs to be processed. You need to either pattern match on it, or use the `?` operator (see https://doc.rust-lang.org/rust-by-example/std/result/question_mark.html). The other `printer.print` line 51 does not have the `?` because it is the last statement of the function and its result is returned straight (pay attention to the missing `;`).

The solution you propose cannot work because the `printer` must be used to print the final output of an action (so it must be used once per action). The printer prints either plain text / table or JSON, depending on the `-o|--output` format. By adding more than one `printer.print` you will mess up the JSON output (you will have many JSON result per action).

Here the different options I had in mind for this issue:

- Aggregate strings "downloading X attachment" together then send it to the final `printer.print`
- Create a new function in the printer that does only print if the output is not JSON (let's say `printer.info`). This way you could call `printer.info` within the loop and keep the final `printer.print`.

The second one seems more appealing to me (and cleaner), but it will require some work. I would understand if you prefer the first version.

Let me know if you have any question!

soywod:
Your PR made me realize that `printer.print` was not the best naming choice. It is confusing. Maybe we could change it to sth more accurate, sth that indicates it needs to be called once. What do you propose? Maybe `printer.print_last`, `printer.print_end`, `printer.return`, `printer.end`… Feel free to propose!

palb91:
Ah! I wasn't ready for this answer… Well, let's try and learn things.
I like the name `printer.return`, it sounds and feel like the END.

soywod:
> Well, let's try and learn things.

This is how himalaya was born :smile: take you time, this is not a priority. Do not hesitate to ask questions if you need :wink:

soywod:
Also, just a detail, but you should start from the `development` branch, not `master` (this last one is only for production). I updated your PR but you also need to update your local branch.

soywod:
> I like the name `printer.return`, it sounds and feel like the END.

I also agree, I hope Rust will allow you to use this keyword. This way we could use `printer.print` for the name of the 2nd function (the one printing only if output not JSON).

palb91:
Well, I cannot use `return` like that, so it will be `printer.end` for the last call and `printer.print` for information, if it's ok for you.
Rust is so different from what I know (Python mostly), there are some concepts that I do not understand well.

So I renamed the function `printer.print` to `printer.end`, then I tried to create a function that just print something. Can I use a simple `println!` call or do you have another existing way to print something? `data.print` seems to only accept `Result` things…

More questions coming soon :D

soywod:
> so it will be `printer.end` for the last call and `printer.print` for information, if it's ok for you.

Yes, sounds great to me!

> Rust is so different from what I know (Python mostly)

Rust is so far from everything (even from C I would say), this ownership thing is really special. The learning curve is hard, but it's worth it :wink:

> Can I use a simple `println!` call or do you have another existing way to print something?

The core concept behind the printer is the `PrinterService`. It is a simple trait that exposes the `print` and the `print_table` functions. Why using a trait? Mostly for testing purposes. It makes the implementations independent from any kind of writer (and not wri*tt*er, as I wrote everywhere in the codebase… shame on me). A concrete example is the `ui/table.rs` (check the tests), you can see that I created a special test printer that uses a `StringWriter`, this way I can test directly the output generated by the app. In production (see the `main.rs`), I use the `StdoutPrinter`, which uses the stdout writer (and so prints directly to the screen).

I tell you all this for you to have a bit more context. To answer your question, I would not use directly `println!`, because this is a macro that writes directly to the stdout. Instead I would pass in parameter a writer (in fact it should be really similar to the `print` implementation of `StdoutPrinter`: https://github.com/soywod/himalaya/blob/585fa77af5058f6fef924eeb52680df03d6a41fc/src/output/printer_service.rs#L25-L31).

> More questions coming soon :D

Feel free to ask, I will try to answer as much as I can!

soywod:
Sorry for the conflict, although it should not be hard to resolve :hand_over_mouth:

soywod:
> Could you tell me why did you close the PR? Is it already fixed somehow or it is just that I was not giving feedback?

I would not dare doing such thing. But I think GitHub automatically deleted the `development` branch when I merged it into `master` for the previous release… I disabled the option so it will not happen again. Sorry!

> I have some changes in my personal and professional life, working on this will be difficult rigth now.

No problem, it is an open source project, come back whenever you can/want. And if you don't it is also fine, someone will take it :wink:

soywod:
I had some free time, so I decided to take care of this feature. I finally opted for `print_str` for the simple message, and `print_struct` for the "final" one, because the result is wrapped into a struct `Response`. `print_table` remains the same. I also fixed this horrible typo `writter` :smile:. Thank you anyway for your initiative, it helped me to improve the printer.

> I have some changes in my personal and professional life

I hope you are doing well!

palb91:
I'm very fine, just busy busy! I hope that you are well too.
I'm glad our discussion was at least useful ^^ I looked at the diff between 212f5e6 and 130ed24, there is much more work that I thought (even considering that you rewrite some part that were not mandatory for this feature).
So… thanks for taking care of this feature :)
Continue with your great work !

PR #299, opened 2022-02-17
Maildir backend

soywod:


src/backends/backend.rs
```diff
@@ -0,0 +1,41 @@
...
+        page: usize,
+    ) -> Result<Box<dyn Envelopes>>;
+    fn add_msg(&mut self, mbox: &str, msg: &[u8], flags: &str) -> Result<Box<dyn ToString>>;
+    fn get_msg(&mut self, mbox: &str, id: &str) -> Result<Msg>;
+    fn copy_msg(&mut self, mbox_src: &str, mbox_dst: &str, ids: &str) -> Result<()>;
+    fn move_msg(&mut self, mbox_src: &str, mbox_dst: &str, ids: &str) -> Result<()>;
+    fn del_msg(&mut self, mbox: &str, ids: &str) -> Result<()>;
+    fn add_flags(&mut self, mbox: &str, ids: &str, flags: &str) -> Result<()>;
+    fn set_flags(&mut self, mbox: &str, ids: &str, flags: &str) -> Result<()>;
+    fn del_flags(&mut self, mbox: &str, ids: &str, flags: &str) -> Result<()>;
```
matthiasbeyer, line 36: So the above functions are all stringly typed.

How about adding a strongly typed interface instead of stringly typed APIs?

soywod: Initially they were typed (for IMAP). But I realized that it did not fit with the Maildir at all (mailboxes does not have attributes nor delimiter, sort criteria are not the same, filters and queries also...). So I decided to stay the most generic possible. I keep str because it is what I receive anyway from the args, and this way the backend implementations are free to map those str into inner types and struct that matches better their need.

But I agree with you, a strong shared typed system could be better, but IMO it should be done once all other backends are implemented. It's not gonna be a easy work. Also, such a typed system will brake the public API (CLI), I prefer to do it once, when the actual backend system is solid enough!

matthiasbeyer: Sure, whatever floats your boat! :laughing: 
I'm just concerned that stringly typed APIs will result in a mess above or below that API... :smile: But as you seem to have a plan for it, go ahead! :+1:

Issue #305, opened 2022-02-23
No tilde expansion in `maildir-dir`


wylfen:
It seems that even though the [example config](https://github.com/soywod/himalaya/wiki/Configuration:config-file#example) has
```toml
maildir-dir = "~/Maildir"
```
shell expansion is not yet supported in the `maildir-dir` variable.

soywod:
Indeed, the tilde (and env vars) are not expanded in the `maildir-dir`. Although it should not be complicated to set up, it has already been done for the `downloads-dir`: https://github.com/soywod/himalaya/blob/158bc86cfaa1ad3ff2b9fb9ec842c7f941e61042/src/config/account_config.rs#L91

soywod:
You can take with great pleasure! Happy to see you back :wink: (don't forget to start from the `development` branch).

soywod:
The bug has been fixed by @TornaxO7 on the `development` branch, it will be shipped with the next release!

PR #307, opened 2022-02-23
305 maildir expansion

TornaxO7:
This should fix https://github.com/soywod/himalaya/issues/305.

Please note that I'm pretty overwhelmed by the new code so I was pretty unsure how to test it.

I'd be happy if some of you guys could quickly try this branch out and give a submission :)

### DIFF
--- a/src/config/account_config.rs
+++ b/src/config/account_config.rs
@@ -2,12 +2,7 @@ use anyhow::{anyhow, Context, Result};
 use lettre::transport::smtp::authentication::Credentials as SmtpCredentials;
 use log::{debug, info, trace};
 use mailparse::MailAddr;
-use std::{
-    env,
-    ffi::OsStr,
-    fs,
-    path::{Path, PathBuf},
-};
+use std::{env, ffi::OsStr, fs, path::PathBuf};
 
 use crate::{config::*, output::run_cmd};
 
@@ -195,11 +190,8 @@ impl<'a> AccountConfig {
                 imap_passwd_cmd: config.imap_passwd_cmd.clone(),
             }),
             DeserializedAccountConfig::Maildir(config) => {
-                let maildir_path_str = config.maildir_dir.to_string_lossy();
-                let expanded_path = shellexpand::full(&maildir_path_str)?.into_owned();
-                let expanded_path = Path::new(&expanded_path);
                 BackendConfig::Maildir(MaildirBackendConfig {
-                    maildir_dir: expanded_path.to_path_buf(),
+                    maildir_dir: shellexpand::full(&config.maildir_dir)?.to_string().into(),
                 })
             }
         };
--- a/src/config/account_config.rs
+++ b/src/config/account_config.rs
@@ -2,7 +2,7 @@ use anyhow::{anyhow, Context, Result};
 use lettre::transport::smtp::authentication::Credentials as SmtpCredentials;
 use log::{debug, info, trace};
 use mailparse::MailAddr;
-use std::{env, ffi::OsStr, fs, path::PathBuf};
+use std::{env, ffi::OsStr, fs, path::{PathBuf, Path}};
 
 use crate::{config::*, output::run_cmd};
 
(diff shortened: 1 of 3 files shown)


soywod:
> Please note that I'm pretty overwhelmed by the new code

Indeed, the config part has been refactored (it is way cleaner than before), and the backend feature changed a lot of thing around the IMAP service. But the core principle is still there :smile: 

> I'd be happy if some of you guys could quickly try this branch out and give a submission :)

I tried your branch and it works well :tada:, but I have few remarks for you (I will put them in a review).

src/config/account_config.rs
```diff
@@ -190,8 +190,14 @@ impl<'a> AccountConfig {
                 imap_passwd_cmd: config.imap_passwd_cmd.clone(),
             }),
             DeserializedAccountConfig::Maildir(config) => {
+                let maildir_path_str = config
+                    .maildir_dir
+                    .to_string_lossy();
+                let expanded_path = shellexpand::full(&maildir_path_str)?
```
soywod, line 196: Just a side note on how the config folder is now organized. You have 3 main files:

- `deserialized_config` which contains the deserialized config file by serde (you can see it as the raw translation of the TOML file)
- `deserialized_account_config` which contains the deserialized struct of an account (from the HashMap located in `deserialized_config`)
- `account_config` which contains the final config of the currently selected account (there is no more config, it has been merged with the account for clarity)

---

`shellexpand::full` expects a `AsRef<str>`, but we have a `PathBuf`. I think the simplest is to change the raw deserialized type of `maildir_dir` (in `deserialized_account_config`, last line) from `PathBuf` to `String`. This way you could simplify all what you did by:

```rust
BackendConfig::Maildir(MaildirBackendConfig {
    maildir_dir: shellexpand::full(&config.maildir_dir)?.to_string().into(),
})
```

soywod (changes requested):

src/config/account_config.rs
```diff
@@ -2,7 +2,7 @@ use anyhow::{anyhow, Context, Result};
 use lettre::transport::smtp::authentication::Credentials as SmtpCredentials;
 use log::{debug, info, trace};
 use mailparse::MailAddr;
-use std::{env, ffi::OsStr, fs, path::PathBuf};
+use std::{env, ffi::OsStr, fs, path::{PathBuf, Path}};
```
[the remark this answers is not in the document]
TornaxO7: Yes you're right.
```toml
required_version = "1.4.37"
```
is this really required? Because my `rustfmt` is on `1.4.37`.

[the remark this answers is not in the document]
soywod: Hmmm not so much. I guess it comes from the default generated config. You can bump it or remove it!

src/config/account_config.rs
TornaxO7: I get the following if I use your suggestion:
```rust
error[E0277]: the trait bound `PathBuf: AsRef<str>` is not satisfied
   --> src/config/account_config.rs:194:52
    |
194 |                     maildir_dir: shellexpand::full(&config.maildir_dir)?.to_string().into(),
    |                                  ----------------- ^^^^^^^^^^^^^^^^^^^ the trait `AsRef<str>` is not implemented for `PathBuf`
    |                                  |
    |                                  required by a bound introduced by this call
    |
    = help: the following implementations were found:
              <PathBuf as AsRef<OsStr>>
              <PathBuf as AsRef<Path>>
note: required by a bound in `full`
   --> /home/tornax/.cargo/registry/src/github.com-1ecc6299db9ec823/shellexpand-2.1.0/src/lib.rs:303:9
    |
303 |     SI: AsRef<str>,
    |         ^^^^^^^^^^ required by this bound in `full`
```

soywod: Because you did not change the type of deserialized `maildir_dir` property (check the `config/deserialized_account_config.rs` file, last line).

TornaxO7: Sorry, I wasn't really aware of your sentence here:
> I think the simplest is to change the raw deserialized type of maildir_dir (in deserialized_account_config, last line) from PathBuf to String.

It's working now. Hm... But I'm wondering if it isn't better if we are still using `Path`, because otherwise there might be some restrictions for some people for their file-/directorynames.

TornaxO7: By the way, thank you for the eplanation of these 3 files :)

soywod: > But I'm wondering if it isn't better if we are still using `Path`, because otherwise there might be some restrictions for some people for their file-/directorynames

I think it does not matter. You only changed the type of the deserialized config, but the type of the final config remains `PathBuf` (note the `.into()` at the end of your changes).

TornaxO7: oh yes, you're rigth :+1:

src/config/account_config.rs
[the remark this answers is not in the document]
TornaxO7: I removed it

soywod:
Thanks :pray:

PR #312, opened 2022-02-27
Config wizard

TornaxO7:
This PR should give himalaya a config wizard like `mutt-wizard`.


Cargo.toml
```diff
@@ -40,6 +40,7 @@ serde_json = "1.0.61"
 shellexpand = "2.1.0"
 termcolor = "1.1"
 terminal_size = "0.1.15"
+thiserror = "1.0.30"
```
soywod, line 43: Why not using the existing anyhow? I'm not so fan to have 2 different error management tools in the same codebase, let's stay consitant.

src/config/mod.rs
```diff
@@ -3,6 +3,7 @@
 pub mod config_args;
 pub mod deserialized_config;
 pub use deserialized_config::*;
+pub mod wizard;
```
soywod, line 6: Did you check if any wizard does not already exist before creating your own? Sth like https://crates.io/crates/requestty or https://crates.io/crates/enquirer or https://crates.io/crates/promptly?

Cargo.toml
TornaxO7: I wanted to use [`thiserror`](https://crates.io/crates/thiserror) to create my own custom errors + messages for the wizard. But since it looks like that I don't need to write our own wizard, due to your links, I think that I'll remove this dependencie then :)

src/config/mod.rs
TornaxO7: I didn't knew that this kind of crates exist! Thank you for pointing them out!

TornaxO7:
So I've a question here @soywod : Do you agree, if I use [`dialoguer`](https://crates.io/crates/dialoguer)? Here are my reasons:
1. [`reqeuestty`](https://crates.io/crates/requestty) seems to be pretty young and I don't know really know how "far" it is.
2. [`enquirer`](https://crates.io/crates/enquirer) seems to be commandline-binary-tool which should be used for bash scripts for example not as a library for rust-project. It's using [`dialoguer`](https://crates.io/crates/dialoguer) underneath.
3. [`promptly`](https://crates.io/crates/promptly) doesn't look as mature as [`dialoguer`](https://crates.io/crates/dialoguer).

What do you think? Please note that I just quickly went through their readme and a little bit through their docs so it might happen, that I skipped some important parts.

soywod:
`dialoguer` looks really good and seems to fit well our needs, let's go for it :slightly_smiling_face:

TornaxO7:
@soywod little question here (I think that I've asked that already once, but I can't find my question anymore):
You wrote the following in your code:
```rust
/// Represents the user account.
#[derive(Debug, Default, Clone)]
pub struct AccountConfig {
    /// Represents the name of the user account.
    pub name: String,
    /// Makes this account the default one.
    pub default: bool,
    /// Represents the display name of the user account.
    pub display_name: String,
    ...
}
```
What's the difference, between `name` and `display_name`? Is it meant to be like this:
```toml
[AccountConfig.name]
name = AccountConfig.display_name
```
?

soywod:
The `name` is the name of the account (the one you put between brackets at the top of every account). See it as the key or the identifier. The `display_name` is the full name of the user that ownes the account.

TornaxO7:
Hi! I've got another questiong: As you mentioned, it'd be pretty nice if we could get the host-data just by looking up the hostname of the e-mail-address. So I'm wondering if we should store the same `.csv` file with the host-information in [`mutt-wizard`](https://github.com/LukeSmithxyz/mutt-wizard/blob/master/share/domains.csv) in himalaya or just download it everytime the user starts the wizard and remove it afterwards. What do you think about this?

TornaxO7:
Hm... I could imagine, that 17.5KB is a lot for some people so I think that downloading it into himalaya is fine.
But now I'm a little bit unsure how we should add those entries into the binary? The most people are probably just downloading the compiled version of himalaya from their package-manager.

I think an optional question to the user, if the wizard should try to figure the host out, can be added, so if the user thinks that downloading a `csv` file, just to get the host-information takes to long, then he/she can say simply "no".

soywod:
> I think an optional question to the user, if the wizard should try to figure the host out on its own, can be added, so if the user thinks that downloading a `csv` file takes to long (just to get the host-information), then he/she can say simply "no".

Yeah, sounds good. Maybe sth like "configuration type: manual | auto".

TornaxO7:
huh, why did you delete the development branch?

TornaxO7:
oh I see, you released a new version

soywod:
> huh, why did you delete the development branch?

The auto-delete branch feature deleted it after having merged `development` into `master` for the release… sorry. I just removed this option!

TornaxO7:
No problem

soywod:
How is it going with this feature? I'm asking because some breaking changes are coming (see https://github.com/soywod/himalaya/discussions/345) and I'm afraid we will reach the same conflicting problems than for the TUI. After the `v0.6.0` we should reach sth quite solid, and it could be the right moment to start big works like the TUI, the GUI and this config wizard.

---

I'm sorry for all those breaking changes. I'm aware that I may also have broken your motivation after the TUI and now the config wizard. Himalaya comes from far (it's been years I'm experimenting sth around), and it was hard to find the place and the direction of this tool. It is only since the `v0.5` and the backend features that I really start to gain confidence and visibility. From now the tool should be more and more solid and should introduce less and less breaking changes. The `v1` should not be so far from the `v0.6`!

soywod:
Yeah we can close. We will come back on it later, when the time comes.

> But for the time being I need to pass my exams first xD

Good luck dude :crossed_fingers: :muscle:

PR #323, opened 2022-03-04
Fix replies

okrplay:
When a message is converted to a reply, the `In-Reply-To`-header is set after the message id has already been cleared. This PR reverses the order to fix this issue.


soywod:
Thank you :pray:

PR #325, opened 2022-03-05
Remove excess whitespace from sig delimiter

polyzen:


soywod:
Nowadays there is no true convention for the signature delimiter. But there is a common practice which consist of using the signature delimiter from the [Usenet news system](https://en.wikipedia.org/wiki/Signature_block#Standard_delimiter). And this one contains a space before the line feed (`-- \n`).

> Most Usenet clients (including, for example, [Mozilla Thunderbird](https://en.wikipedia.org/wiki/Mozilla_Thunderbird)) will recognize the “dash dash space” delimiter in a news article and will cut off the signature below it when inserting a quote of the original message into the composition window for a reply.

If you personally do not want the space, you can define your own one in the configuration (check the [wiki](https://github.com/soywod/himalaya/wiki/Configuration:config-file) about `signature-delimiter`).

polyzen:
Did not notice Thunderbird included the space o.O Confused why that's a thing.

soywod:
> Confused why that's a thing.

I should admit it is quite useful to detect signature and to remove them when replying/forwarding an email. But it's just a detail!

PR #328, opened 2022-03-06
Minor consistency fix

polyzen:


soywod:
Nice catch :laughing: thank you!

PR #357, opened 2022-04-06
Default Content-Type to text/plain for not multipart messages

fabrixxm:
Parse body of messages without subparts as text/plain if Content-Type header is not set.

I've found emails sent by `git send-email` that do not have 'Content-Type' header, therefore body is discarded and nothing is printed with `himalaya read`.


soywod:
First of all, thank you for your contribution!

I think we need to be more precise. If an attachment does not contain
any `Content-Type` header, it will be considered as a `text/plain`
(imagine an image, or sth that does not display well as a text
part).

What about narrowing a bit more the condition? Sth like: if a message
has only one part AND it does not contain any `Content-Type`, then take
it as a `text/plain`? Btw, do git emails contain only one part?

fabrixxm:
For what I understand of the code but I can be wrong) my patch is already in a condition "a message
has only one part" (or "is not a multipart message" ?  )

from line 82:
```rust
     if parsed_mail.subparts.is_empty() {
        let cdisp = parsed_mail.get_content_disposition();
        match cdisp.disposition {
            mailparse::DispositionType::Attachment => {
                 // [...]
            }
            // TODO: manage other use cases
            _ => {
                let ctype = parsed_mail.get_headers()
                    .get_first_value("content-type")
                    .unwrap_or_else(|| String::from("text/plain"));
```

>  Btw, do git emails contain only one part?

As far as I know, git-send-email send a non-multipart email. Here one of the email that made me send this patch:

https://lists.sr.ht/~valhalla/lesana-devel/%3C20220111160733.51730-1-fabrixxm%40kirgroup.net%3E/raw

soywod:
> For what I understand of the code but I can be wrong) my patch is
> already in a condition "a message has only one part" (or "is not a
> multipart message" ?  )

The function that wraps the condition is kind of recursive. The
condition means that you are in a single part, but this single part
could be part of a multipart. To check if a message contains only one
part, we need to add a condition before calling this recursive function
(have a look at `Parts::from_parsed_mail`).

> As far as I know, git-send-email send a non-multipart email. Here one
> of the email that made me send this patch:

Great, so the condition at the top level should be enough :)

fabrixxm:
> For what I understand of the code but I can be wrong) my patch is already in a condition "a message has only one part" (or "is not a multipart message" ? )
> The function that wraps the condition is kind of recursive. The condition means that you are in a single part, but this single part could be part of a multipart. To check if a message contains only one part, we need to add a condition before calling this recursive function (have a look at `Parts::from_parsed_mail`).

Oh, I didn't see that. I'll have a look and I'll send an updated patch, thank you!

soywod:
Looks great, thank you :tada:

soywod:
> For what I understand of the code but I can be wrong) my patch is
> already in a condition "a message has only one part" (or "is not a
> multipart message" ?  )

The function that wraps the condition is kind of recursive. The
condition means that you are in a single part, but this single part
could be part of a multipart. To check if a message contains only one
part, we need to add a condition before calling this recursive function
(have a look at `Parts::from_parsed_mail`).

> As far as I know, git-send-email send a non-multipart email. Here one
> of the email that made me send this patch:

Great, so the condition at the top level should be enough :)

PR #358, opened 2022-04-07
renaming the `Print` trait to `Printable`

TornaxO7:
Just a little renaming which sounds a little bit better in my opinion.


TornaxO7:
wait what? Conflict files?

soywod:
Thank you for you proposition. However, I prefer to keep the original
version for 2 reasons:

- `Printable` conflicts with the other trait `PrintTable`.

- Rust standard traits do not use this convention. Have a look at this
  discussion: https://github.com/rust-lang/api-guidelines/discussions/28

TornaxO7:
Oh interesting, thank you for the link to the discussion!

PR #361, opened 2022-04-09
add `tpl_args` for `write` subcommand

illfygli:
Added this in my fork so that `himalaya write` can be used as a `mailto:`handler.

Happy to make any changes needed for merge if you think this addition makes sense.


soywod:
> Added this in my fork so that `himalaya write` can be used as a
> `mailto:`handler.

Thanks a lot for your contribution! I thought about this not so long
time ago and it is a very good addition.

> Happy to make any changes needed for merge if you think this addition
> makes sense.

I see that you added the `--to` arg, but in fact we could add other args
like `--cc`, `--subject` etc. Have a look at the `TplOverride` struct in
`tpl_args`, it does exactly this.

illfygli:
Thanks for the hint, I gave it a go. :)

cli/src/msg/msg_entity.rs
```diff
@@ -425,6 +426,11 @@ impl Msg {
         self
     }
 
+    pub fn to(mut self, to: Vec<&str>) -> Result<Self> {
```
soywod, line 429: This function can be removed if it is not used any more.

soywod (changes requested):
Sounds really great! I just have a last comment about dead code, and we can merge it.

soywod:
I also noticed that you targeted the wrong branch (`master`), I just switched it to `development`.

illfygli:
Now I discovered that it already works as a mailto-handler with just `himalaya` and no subcommand. :smile: 

But good to have this option as well, I think.

soywod:
> Now I discovered that it already works as a mailto-handler with just
> `himalaya` and no subcommand.

Ah yes indeed :) I realized that the mailto support is not documented at
all. I will update the README.

> But good to have this option as well, I think.

Yes anyway. Thanks a lot!

PR #362, opened 2022-04-11
Check the global config for notify-cmd

JCapucho:
This brings it in line with how notify-query works and how the wiki
defines it to also be a global option.


soywod:
Thank you!

PR #375, opened 2022-04-30
Always reset colors settings on the output stream after writing

robertgzr:
This is according to:
https://docs.rs/termcolor/1.1.2/termcolor/#example-using-standardstream
Not resetting the color settings on the stream will leak the style to
the shell otherwise.
This can be observed when listing mailboxes prior to this patch.

Signed-off-by: Robert Günzler <r@gnzler.io>


soywod:
Thank you :tada:

PR #379, opened 2022-05-12
`read -t html` fallback to plain content if no html is found

### LINKED ISSUE #352: read -t html causes plain text emails to be blank
Disclaimer: I don't know if this is a himalaya thing or an IMAP thing - but either way, it would be nice for himalaya to be able to get around it

I _generally_ want my emails to be read as HTML as I'm piping them into `w3m`, but this doesn't work for plain text emails - it just displays an empty message instead

```
❯ himalaya read 232
hello this is a test email


❯ himalaya read 232 -t html


```

Ideally there would be some way of having himalaya return HTML where possible, otherwise fall back to plain text

fabrixxm:
Still output valid html. Should fix #352

### DIFF
--- a/cli/src/msg/msg_entity.rs
+++ b/cli/src/msg/msg_entity.rs
@@ -167,7 +167,16 @@ impl Msg {
     /// and html text parts.
     pub fn fold_text_parts(&self, text_mime: &str) -> String {
         if text_mime == "html" {
-            self.fold_text_html_parts()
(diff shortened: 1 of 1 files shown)


fabrixxm:
nvm. it doesn't work correctly.

soywod:
Thanks for the try anyway :) the condition should be higher in the code. I will try to do sth this weekend and let you know.

PR #383, opened 2022-05-18
vim: msg_id is a string

robertgzr:
my msg_id's look like: b85, d39, ae0, etc.
and I get this error when using the vim plugin:

	Error: cannot find maildir message by short hash "0" at "/path/to/my/INBOX"
	Caused by:
	0: cannot find maildir message id from short hash "0"
	1: the short hash "0" matches more than one hash: 030598120934103c456ce08338886728, 06edb10a55efb89de45d8560aee33c8e

Signed-off-by: Robert Günzler <r@gnzler.io>

### DIFF
--- a/vim/autoload/himalaya/msg.vim
+++ b/vim/autoload/himalaya/msg.vim
@@ -3,7 +3,7 @@ let s:trim = function("himalaya#shared#utils#trim")
 let s:cli = function("himalaya#shared#cli#call")
 let s:plain_req = function("himalaya#request#plain")
 
-let s:msg_id = 0
+let s:msg_id = ""
 let s:draft = ""
 let s:attachment_paths = []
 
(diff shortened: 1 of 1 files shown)


vim/autoload/himalaya/msg.vim
```diff
@@ -375,7 +376,7 @@ function! s:bufwidth()
 endfunction
 
 function! s:get_msg_id(line)
-    return matchstr(a:line, '[0-9]*')
+  return matchstr(a:line, '[0-9a-z]*')
```
soywod, line 379: In fact it should be `[0-9a-fA-F]*`, see https://datatracker.ietf.org/doc/html/rfc4122#section-3.

soywod (changes requested):
Thanks a lot for you contribution! Indeed, Maildir message ids are UUID strings and the Vim plugin was not up-to-date. Did you try with a regular IMAP account to see if it is backward compatible? I just left one comment, we can merge straight after.

robertgzr:
> Did you try with a regular IMAP account to see if your patch is backward compatible?

just did. works beautifully :)

soywod (approved):

soywod:
Thanks a lot :pray:

PR #385, opened 2022-05-28
Update gitignore

TornaxO7:
Those are just some small adddings into the `gitignore` which don't hurt I hope.
Also I did `cargo fmt` so I think that those code-changes should be fine?


.gitignore
```diff
@@ -1,10 +1,48 @@
 # Cargo build directory
-/target
+target/
+debug/
+Carge.lock
```
soywod, line 4: I think we should move this one to the `cli` folder (plus there is a typo, you typed `Carge` instead of `Cargo` :smile:). If I understand well, the lock file should be present in lib folders but absent from binary folders, see https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html.

.gitignore
```diff
@@ -1,10 +1,48 @@
...
+
+# Crashlytics plugin (for Android Studio and IntelliJ)
+com_crashlytics_export_strings.xml
+crashlytics.properties
+crashlytics-build.properties
+fabric.properties
+
+## Others
+.metadata/
+himalaya.iml
```
soywod, line 48: What are those two last for?

soywod (commented):
Thanks a lot for this initiative :pray: it is cleaner to use gitignore patterns from https://github.com/github/gitignore. I just left two comments!

.gitignore
TornaxO7: > If I understand well, the lock file should be present in lib folders but absent from binary folders, 

so I should move the whole `.gitignore` to the `cli` folder?

.gitignore
TornaxO7: `.metadata` was due to my neovim-setup because the java-lsp seems to startup (for whatever reason) and I couldn't fix it yet. This can be probably removed since this case just comes up due to me or can I let it in for the time being?

`himalaya.iml`: I don't really know.... but it seems like that it was generated by the java-lsp again (ugh). I'll remove this

.gitignore
soywod: Mmmh no I think we should let it at the root folder, but we should just add a specific one for `Cargo.lock` in the `cli` folder.

TornaxO7: should be done now

cli/.gitignore
```diff
@@ -0,0 +1,47 @@
+# Cargo build directory
+target/
+debug/
+Cargo.lock
```
soywod, line 4: I can put only this line, the rest should be caught by the root .gitignore (can you confirm?)

soywod (changes requested):
Last detail and we can merge :muscle:

.gitignore
soywod: I guess you can remove both of them so, thanks for the details :ok_hand:

cli/.gitignore
TornaxO7: Done and yes I just did `git rm -rf --chached .` and then `git add --all` and `git commit` and this popped up:
```
On branch update-gitignore
Your branch is up to date with 'origin/update-gitignore'.

nothing to commit, working tree clean
```
so looks good to me :+1:

soywod:
Thanks :pray:

PR #386, opened 2022-05-28
Add rust toolchain.toml

TornaxO7:
Adding the [rust-toolchain.toml](https://rust-lang.github.io/rustup/concepts/toolchains.html) so everyone who develops on this has the some tools (`rustc`, `formatter` and so on).
I inserted the version which I'm using. You might change it to your needs before merging this.


soywod:
I like the idea, but it conflicts with Nix. Nix provides a fully functional dev env with already defined rust compiler version, cargo, formater etc. In fact it acts a bit like rustup in a more generic way. I checked back the [overlay](https://github.com/oxalica/rust-overlay) we are using and I found this:

> If you already have a [rust-toolchain file for rustup](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file), you can  simply use fromRustupToolchainFile to get the customized toolchain derivation.
> ```nix
> rust-bin.fromRustupToolchainFile ./rust-toolchain
> ```

So in fact it could be possible to add this `rust-toolchain.toml` file and make it the source of truth. If you use Nix then it will take the versions included in this file, and if you don't you can use your own rustup to use the same versions. Do you allow me to push on your branch some Nix changes? Also I would not like to go for the nightly version, I rather prefer to stay on the stable channel (as it is already the case with Nix right now).

TornaxO7:
> Do you allow me to push on your branch some Nix changes?

aye!

> Also I would not like to go for the nightly version, I prefer to stay on the stable channel (as it is already the case with Nix right now).

Then this toolchain would be very useful since I'm always on nightly xD

soywod:
I pushed two commits, you can have a look. If it's okey for you then we can merge! I opened an issue https://github.com/oxalica/rust-overlay/issues/88 to know if it is normal or not to still specify `rustfmt` and `rust-analyzer` in the `flake.nix` despite the fact that they are specified in the `rust-toolchain.toml`.

rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "1.58.1"
+components = ["cargo", "rustc", "rustfmt", "rust-analysis"]
```
TornaxO7, line 3: @soywod why did you add this line? Do you only want to have those components?

TornaxO7 (commented):
1 open request/question

rust-toolchain.toml
soywod: Well there is not that much available https://rust-lang.github.io/rustup/concepts/components.html, do you see any other components to add?

soywod: We can also remove this line but I do not know which component is added by default?

TornaxO7: hm... I think that all components are used per default and if I think about it again, then we can leave this line there, except adding the `clippy` component! Some people might use it and I think that it's pretty useful.

TornaxO7: I think that this is fine now

soywod:
Thanks again :pray: 

---

Happy to see you here :slightly_smiling_face: I hope you are doing well. How your studies arg going?

TornaxO7:
> Happy to see you here slightly_smiling_face

Thanks ^^

> I hope you are doing well. How your studies arg going?

Yes I'm doing well, thank you for asking me that :) Well, I passed my exams last semester but this semester is more stressful... anyhow I practical stuff instead of doing only theoretical stuff, so I'm coming for some time back here :)

soywod:
> I want to do some practical stuff as well instead of doing only theoretical stuff, so I'm coming for some time back here :)

Welcome back, the door is always open :wink: 

Himalaya may receive some financial help from government, which would allow me to allocated more time on this project. I will announce it publicly if it happens!

TornaxO7:
> Himalaya may receive some financial help from government,

holy cow this sounds awesome! Would it also be possible to public your discussion or at least a summary? I'm veeeeeeery interested how the government got interested in it c(^-^)c

soywod:
Yes ofc, but I first want to wait if they gonna follow me or not. In fact the European Commission has some available funds for FOSS related to security, and it looks like email are part of it. I passed the first round but there is two more rounds so… nothing is done ^^

PR #404, opened 2022-08-10
vim: update new message/draft behavior

benbusby:
Updated himalaya#msg#write:
  - New buffer opens in split view (mirrors behavior of opening a
    message from the mailbox list view)
  - Function accepts an optional argument for populating the buffer with
    existing content, rather than relying on the default msg template
    (see updated "cancel" behavior below)

Updated himalaya#msg#draft_handle:
  - "c" option (to cancel exiting the new message buffer) now passes the
    existing buffer content back to himalaya#msg#write. This is because
    Vim does not allow event interruption. Without the ability to stop
    the buffer closing event, message content will be lost unless
    preserved and reinserted. With these changes, the "c" option now
    seems -- for the user -- as if the request to close the buffer was
    truly cancelled.


soywod:
Awesome, thank you :tada:

PR #406, opened 2022-08-16
vim: match supported commands to available shortcuts

benbusby:
This expands the list of supported "full name" commands (i.e.
`Himalaya<Action>`) to match the list of available actions supported by
the default keyboard shortcut bindings for Himalaya. This is mostly
preferential, but it's sometimes easier and faster to invoke vim
commands based on the name of the action you're seeking to perform,
rather than trying to remember a specific keyboard binding.

___

Also, thanks for making Himalaya -- I've really been enjoying using it!


soywod:
Thank you :tada:

PR #414, opened 2022-10-11
Add repology badge

jubalh:
As more distributions start to package himalaya I think this information might be useful.


soywod:
Nice one, thank you :pray:

PR #432, opened 2022-12-15
Configuration wizard

kmaasrud:
Another try at #312 with the new API, namely greeting the user with a wizard when no configuration can be found.

I'm aiming at making the wizard flexible wrt. the configuration values, but I hope this PR gives us a chance to properly evaluate the config structure. I feel that we should be able to streamline it a bit, but I'm not sure how (yet.)


soywod:
Linked issue: https://todo.sr.ht/~soywod/himalaya/19.

soywod (changes requested):
Really nice job :muscle: I just left some minor comments. Some of the are related to formatting, I am wondering if it cannot be fixed by updating the [`rustfmt`](https://rust-lang.github.io/rustfmt) config. Some comment are related to printing (with `println!`), I think we should use the `Printer` instead. If I do not mistake, `println!` is just a macro for printing stuff in the `stdout`, whereas the `Printer` prints in a writer (which is really useful for testing, see `ui/table/table.rs` tests).

I will test the feature and potentially review a second time if I find some issues. Thanks a lot for your work :pray:

PR #446, opened 2023-04-29
 Removed the `docs/` from the link to the documentation

qbi:
It seems there is no directory `docs/`, so currently the link shows an 404 page. I removed it that the link is correct.


soywod:
The fix is already on `develop`, changes will be visible within the next release. Thank you!

PR #449, opened 2023-06-04
update sample config to v0.8

icp1994:
I had to update my personal config after the recent update so I thought I'd update the sample config here as well. But I don't actually use Gmail with Himalaya so hasn't tested this particular config.


soywod:
Nice catch, I forgot to update the config sample. Thank you!

PR #451, opened 2023-06-19
fixes the immediate fallout from v0.8.0 build tarball change

yeled:


soywod:
> shall we delete the install.sh then? :)

Exactly!

PR #456, opened 2023-10-21
Fix editor launch without an attached TTY

michael2to3:
An issue was identified where editors like `vim`, `vi`, or `nvim` might not display correctly when invoked from the program, likely due to the absence of an attached TTY.

Changes:
- Transitioned from `run` to `spawn` for improved control over the launched process.
- Implemented the `wait` method on the spawned process to ensure the editor session completes before continuing.

Before: 
image

After:
image

### DIFF
--- a/src/ui/editor.rs
+++ b/src/ui/editor.rs
@@ -6,8 +6,8 @@ use email::{
     sender::Sender,
 };
 use log::debug;
+use std::process::Command;
 use mml::MmlCompilerBuilder;
-use process::Cmd;
 use std::{env, fs};
 
 use crate::{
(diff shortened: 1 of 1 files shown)


soywod:
Thank you for the fix. Just for the context: `process-lib` is part of the Pimalaya project and it is used by all other subprojects, so it would be better to fix directly `process-lib` rather than Himalaya CLI. There is an open issue for this purpose: <https://todo.sr.ht/~soywod/pimalaya/150>, I planned to take care of it very soon. If you want, you can send a patch, see <https://git.sr.ht/~soywod/pimalaya#contributing> and <https://git.sr.ht/~soywod/pimalaya/tree/master/item/process/src/lib.rs>.

PR #457, opened 2023-10-21
Editor opening subprocess now works properly with in-terminal editors.

d2718:
However the `process-lib` crate's [`Cmd`](https://docs.rs/process-lib/latest/process/enum.Cmd.html) runs subprocesses doesn't play nice with in-terminal editors like nano or Helix. I replaced this with an explicit use of [`tokio::process::Command`](https://docs.rs/tokio/latest/tokio/process/struct.Command.html#method.output); the result seems to work properly with both in-terminal and graphical editors.

### DIFF
--- a/src/ui/editor.rs
+++ b/src/ui/editor.rs
@@ -1,4 +1,4 @@
-use anyhow::{Context, Result};
+use anyhow::{ensure, Context, Result};
 use email::{
     account::AccountConfig,
     backend::Backend,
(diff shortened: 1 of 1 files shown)


d2718:
Okay, this is embarrassing; I can't get this to work anymore. Please stand by.

michael2to3:
Hi. You can check my pr it solves the same problem but in a different way

soywod:
Thank you for the fix, I appreciate the initiative! We should fix `process-lib` instead of Himalaya CLI, see my comment <https://github.com/soywod/himalaya/pull/456#issuecomment-1791960641>. A fix will come very soon.

PR #463, opened 2024-01-01
allow account.sendmail when sendmail feature is on

w3irdrobot:
when default features are turned off and the `sendmail` feature isn't enabled, it doesn't build. this fixes that.


soywod:
You were faster than me, I introduced the cargo features `maildir` and `sendmail` but I had no time yet to make them effective. Thank you!

PR #464, opened 2024-01-01
fix ambiguous email export

w3irdrobot:
i'm working on updating the nixos module definitions for himalaya to be able to build the most recent 1.0.0-beta. when building with the next version of rust about to be available in nixpkgs (1.70.0), this ambiguous export was found and preventing building. the version of rust used in this repo's nix environment allows the build to proceed. locally, i updated the `flake.nix` to more closely emulate the build environment of nixpkgs. i can commit that as well in this PR if desired.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -22,5 +22,4 @@ pub mod smtp;
 pub mod ui;
 
 #[doc(inline)]
-// pub use email::{envelope, flag, message, template};
-pub use email::{envelope, flag, message};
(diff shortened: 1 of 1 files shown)


soywod:
> i'm working on updating the nixos module definitions for himalaya to be able to build the most recent 1.0.0-beta

I appreciate the initiative, although it's already on its way, see <https://github.com/NixOS/nixpkgs/pull/278088>.

> i can commit that as well in this PR if desired.

Sure, open the PR so we can discuss on!

PR #504, opened 2024-11-20
Draft: Long version & nix flakes

KoviRobi:
Re: https://github.com/pimalaya/himalaya/issues/502

Will put more words here, just have to head off, but got something reasonable.


build.rs
```diff
@@ -3,11 +3,24 @@ use std::env;
 use git2::Repository;
 
 fn main() {
-    let repo = Repository::open(".").expect("should open git repository");
-    let head = repo.head().expect("should get HEAD");
-    let branch = head.shorthand().expect("should get branch name");
-    let commit = head.peel_to_commit().expect("should get HEAD commit");
-    let rev = commit.id().to_string();
+    let branch = if let Ok(describe) = env::var("GIT_DESCRIBE") {
```
soywod, line 6: This will fail if you build locally using Nix with overriding output attributes (as you did for the CI).

soywod (changes requested):
Interesting approach. My only concern is when building locally using Nix (see my comment). Plus it forces users to manually override (whereas it should definitely be automatic). My fix works everywhere, but I don't have the git branch info anymore. Could be great to find a in-between solution.

.github/workflows/pre-release.yml
```diff
@@ -10,15 +10,20 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - target: x86_64-linux
+          - host: x86_64-linux
+            target: x86_64-linux
```
soywod, line 14: This is definitely sth I wanted to do, I will keep it for https://github.com/pimalaya/core/issues/22.

build.rs
KoviRobi: The CI fail was more about the openssh, I think
https://github.com/KoviRobi/himalaya/actions/runs/11935131223/job/33265788614#step:5:1618
the others worked? So rebasing on top of your changes it should now work.

I have given the environment variables preference, that way you don't have to remove/move the `.git` dir/file.

It works both locally without overriding:
```console
$ nix build .
$ ./result/bin/himalaya --version
himalaya v1.0.0 on linux musl x86_64, git flake-a0485ff rev a0485ff8d18482499bde88e177ff8d989b0d2e5e
```
and with overriding
```
$ nix build --expr "(builtins.getFlake \"git+file://${PWD}?shallow=1&rev=$(git rev-parse HEAD)\").outputs.packages.x86_64-linux.default.overrideAttrs (old: { cargo_build_options = old.cargo_build_options or [] ++ [\"-F\" \"oauth2\"];  GIT_DESCRIBE = \"$(git describe --always HEAD)\"; })"
```

I have also managed to override it in a flake, including cargo features:
https://github.com/KoviRobi/nixos-config/commit/b63057268b5fa39b16b14d987ed86232c8599034#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R86-R100 though some of this is naersk weirdness I'm not sure I got correct.

flake.nix
```diff
@@ -134,7 +134,8 @@
             nativeBuildInputs = with pkgs; [ pkg-config ];
             CARGO_BUILD_TARGET = targetConfig.rustTarget;
             CARGO_BUILD_RUSTFLAGS = [ "-Ctarget-feature=+crt-static" ];
-            GIT_REV = self.rev or self.dirtyRev or "dirty";
+            GIT_REV = self.rev or self.dirtyRev or "unknown-rev";
```
KoviRobi, line 137: Ah this change isn't necessary, only the addition below -- I just had a different name and during the merge conflict I kept the wrong one.

KoviRobi:
Of course this whole thing is just a suggestion, of how I had it in my mind, I don't feel strongly about it. My main reason is so that if you do a tag/release in github, it should presumably build a proper tagged version for you, hence passing in the `git describe` in the github workflow.

soywod:
Thank you for your suggestion. I wanted to push a commit on your branch before merging but I  was not able to, so I merged then applied my commit https://github.com/pimalaya/himalaya/commit/5c4b03474e7bbcf71f4739f11d56c0cbb8c06890. I basically changed few details here and there, and added cargo features to the long help.

KoviRobi:
Ah interesting, that wasn't intentional, I thought I had clicked "allow edit by maintainers", my bad

PR #506, opened 2024-11-27
Nix refactor

### LINKED ISSUE #500: Cannot install on armv7 (0.9.0)
```
pi@REDACTED:~/$ cargo install himalaya
    Updating crates.io index
  Downloaded himalaya v0.9.0
  Downloaded 1 crate (89.4 KB) in 0.22s
  Installing himalaya v0.9.0
    Updating crates.io index
error: failed to compile `himalaya v0.9.0`, intermediate artifacts can be found at `/tmp/cargo-installx8p5Eq`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

Caused by:
  failed to select a version for the requirement `keyring-lib = "=0.1.0"`
```

soywod:
There is too much duplicated Nix code among Pimalaya projects and Nixpkgs.

This PR brings the first milestones:

- Move Nix flake logic to regular `default.nix` and `shell.nix`. This way dependencies and cargo features can be easily overriden by `--arg` and `--argstr`. Nix flake can still be used like before.
- Override the existing nixpkgs derivation instead of re-building a Rust package with `naersk` or `crane`. This way the source of truth resides in and only in nixpkgs. `package.nix` contains the final package that will be released inside nixpkgs, it should disappear as soon as possible.

Aside from Nix, this PR also:

- Initiate a new GitHub Action to build a release on demand, with customizable cargo features.
- Bump Rust to `1.82`.

Related to #504 #505 https://github.com/pimalaya/core/issues/22.
Fixes #500.


soywod:
@KoviRobi I would love your opinion about this proposal.

KoviRobi:
Sorry, I wanted to get back to you, but didn't have the time/energy.

> EDIT: I may merge the PR before your opinion, it does not mean I do not care about it (we can always adjust afterwards), I just need to validate the common Nix repo for Pimalaya and go forward.

Very reasonable, sorry sometimes life gets in the way, plus this is hobby/interest for me to get a decent mail setup so often slow to reply.

I think it looks good, I see you've managed to remove duplication! I'm not a nix expert btw, just saw some issues I could help with so did :)

And I do really like the whole pimalaya project you have going on, I was looking to have a better e-mail setup than outlook web app, especially as at work we use gerrit which sends a lot of emails so I'd need some more complex filtering than outlook can do (maybe something with notmuch and taskwarrior?). But wanted to say that realistically I will only be sporadically committing/contributing, and I think you've got a good approach, I'm also of the view that things can be fixed later too, don't want pull requests to drag on forever.

Sorry I'm rambling a bit, I'm just not sure I/we as a community have found the best way to do open source software, and ever since the XZ utils, it's been on the back of my mind, and keep just hoping to find ways in which my contributions will be a net positive rather than negative, and ways to help avoid burnout of maintainers for software I use.

soywod:
No need to be sorry, we all do what we can with what we have. I am really thankful to your initiative, the Nix topic was sth I was postponing for too long, and you initiated the changes.

> maybe something with notmuch and taskwarrior?

If you stay tuned, you may see a task & time manager joining the Pimalaya family next year! And should integrate well with other projects. It will be a refactor of the existing [Unfog](https://github.com/soywod/unfog) project of mine, using Rust and Pimalaya libs. You should also see contact and calendar tools.

PR #520, opened 2024-12-10
s/boby/body/ (fix spelling in README)

Gumnos:


soywod:
Nice catch, thank you!

PR #540, opened 2025-01-11
docs: link to conventionalcommits.org instead of their repo

ajanvrin:
Standardizing commit style is welcome, but I was not aware of what Conventional Commits are, and their repository does not describe what the commits are meant to look like. Linking to the summary is IMHO more convenient.

Uh, for clarity: I was just curious about what Conventional Commits are, I am far from being familiar enough with the code base to contribute anything substantial other than hopefully helpful bug reports.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -66,4 +66,4 @@ secret-lib.path = "/path/to/core/secret"
 
 ## Commit style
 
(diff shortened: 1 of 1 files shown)


ajanvrin (commented):
aaaaaaaaaand I forgot the mandatory blank line between the description and the optional body.
epic fail

soywod:
You are right, it is definitely better. Thank you!

PR #558, opened 2025-02-09
build: align package.nix with nixpkgs version; verbatim copy

blaggacao:
This aims to reduce confusion for people who more closely want to track the pimalaya project in their nixos convfigurations with overlays.


soywod:
Now that the derivation is up-to-date on nixpkgs, I would rather delete `package.nix` and update `default.nix` [to use the nixpkgs derivation instead of a local `package.nix`](https://github.com/pimalaya/himalaya/blob/26feecd80ac66f398d8b3333c64aa4a0f9053a79/default.nix#L18). I will give it a shot whenever I can. Thanks for pointing it out!

blaggacao:
> Now that the derivation is up-to-date on nixpkgs, I would rather delete package.nix and update default.nix [to use the nixpkgs derivation instead of a local package.nix](https://github.com/pimalaya/himalaya/blob/26feecd80ac66f398d8b3333c64aa4a0f9053a79/default.nix#L18).

That seems like a good idea, indeed! &mdash; I got my setup working, but I still needed to research a while into how building all the things I needed. Having nixpkgs be the authoritative origin for v1+ packages seems like equally useful.

PR #564, opened 2025-02-20
Fix gmail doc example

thornycrackers:
When attempting to use the gmail provider example I was getting errors pointing to the line `backend.type.host`. I compared the configuration to the other examples and found none of them had the additional `type` in their other settings. This worked for me so I'm guessing it's just a typo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,11 +295,11 @@ You can also manually edit your own configuration, from scratch:
   folder.aliases.trash = "[Gmail]/Trash"
 
   backend.type = "imap"
-  backend.type.host = "imap.gmail.com"
-  backend.type.port = 993
-  backend.type.login = "example@gmail.com"
-  backend.type.auth.type = "password"
-  backend.type.auth.raw = "*****"
+  backend.host = "imap.gmail.com"
+  backend.port = 993
+  backend.login = "example@gmail.com"
+  backend.auth.type = "password"
+  backend.auth.raw = "*****"
 
   message.send.backend.type = "smtp"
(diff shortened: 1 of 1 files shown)


thornycrackers:
```bash
󰘧  himalaya --version
himalaya v1.1.0 +wizard +pgp-commands +maildir +smtp +sendmail +imap
build: linux gnu x86_64
git: unknown, rev unknown

󰘧  himalaya envelope list
Error:
   0: cannot parse config file at /home/thorny/.config/himalaya/config.toml
   1: TOML parse error at line 13, column 1
   1:    |
   1: 13 | backend.type.host = "imap.gmail.com"
   1:    | ^
   1: dotted key `backend.type` attempted to extend non-table type (string)
   1:

Note: Run with --debug to enable logs with spantrace.
Note: Run with --trace to enable verbose logs with backtrace.
```

That's with `type.` included in each line and looking like this:

```
backend.type = "imap"
backend.type.host = "<host>"
backend.type.port = <port>
backend.type.login = "<email>"
backend.type.auth.type = "<type>"
backend.type.auth.raw = "<raw>"
```

Then I change it to:

```
backend.type = "imap"
backend.host = "<host>"
backend.port = <port>
backend.login = "<email>"
backend.auth.type = "<type>"
backend.auth.raw = "<raw>"
```

And everything works. Your README.md has a bad example in it that is not consistent with [config.sample.toml](https://github.com/pimalaya/himalaya/blob/master/config.sample.toml#L357-L382)

PR #636, opened 2026-02-26
fix: parse multipart BODYSTRUCTURE with spaces between body parts

aseedb:
## Summary

- Fix IMAP BODYSTRUCTURE parsing failure when servers send spaces between body parts in multipart structures (e.g., `(body1) (body2) "alternative"` instead of `(body1)(body2) "alternative"`)
- Vendor imap-codec v2.0.0-alpha.7 with a patched `body_type_mpart_limited` parser that accepts optional spaces between body parts
- Add test covering spaces between body parts with and without extension data

## Details

Many IMAP servers (e.g. Stalwart, Exchange) send spaces between body parts in multipart BODYSTRUCTURE responses. While RFC 3501's ABNF `1*body` implies no separator, this is extremely common in practice.

The imap-codec parser used `many1(body(...))` which required body parts to be directly adjacent. When spaces appeared, the parser stopped after the first body, then failed to find the media subtype, causing the entire FETCH response to be discarded as "skipping invalid fetch".

The fix replaces `many1(body(...))` with a manual loop that parses the first body, then tries additional bodies with optional spaces before each one, restoring the parse position if a space is consumed but no body follows.

## Test plan

- [x] All 125 existing imap-codec unit tests pass
- [x] All 45 imap-codec integration tests pass
- [x] New test `test_parse_body_mpart_with_spaces_between_parts` covers the fix
- [x] Tested with real emails from Exchange/Outlook via Stalwart IMAP server — all 8 messages now display correctly (previously only 1 of 8 was shown)


soywod:
Thanks for the fix, but I believe you published the PR on the wrong repository. Open your PR there instead: https://github.com/duesee/imap-codec.

aseedb:
Yes this makes absolute sense! Will do

PR #642, opened 2026-03-10
Fix references of `imap.sasl.login.password` and `smpt.sasl.login.password`

knownasnaffy:


config.sample.toml
```diff
@@ -94,7 +94,7 @@ smtp.sasl.mechanisms = ["plain", "login"]
 # SASL LOGIN
 # https://datatracker.ietf.org/doc/html/draft-murchison-sasl-login-00
 smtp.sasl.login.username = "username"
-smtp.sasl.plain.password.command = ["mimosa", "password", "read", "example"]
+smtp.sasl.login.password.command = ["mimosa", "password", "read", "example"]
 #smtp.sasl.plain.password.raw = "***"
```
soywod:
```suggestion
#smtp.sasl.login.password.raw = "***"
```

soywod (changes requested):
Nice catch, thank you. I found another one just below, then I will merge your PR. 

EDIT: in fact I was able to apply changes myself on your branch!

config.sample.toml
knownasnaffy: Right, I missed those :P

PR #655, opened 2026-03-20
bump rust to 1.90

yobert:
Rustup defaults to an old version without this fix. Thanks!


soywod:
The toolchain file will be removed from the `v2`, as it serves no purpose. Users should be free to use any version they need, as long as it matches the minimum version recommended by Himalaya CLI. Thanks!

PR #661, opened 2026-04-02
docs(readme): fix v1.1 account command examples

artemgetmann:
## Summary

- update the top-level README example to use subcommand-local account selection
- replace the stale Gmail setup command with the current `account configure <name>` form

## Why

The README still includes two command examples that drifted from the current v1.1 CLI:

- `himalaya envelope list --account ...`
- `himalaya configure -a gmail`

The current CLI expects:

- account selection on the concrete subcommand, e.g. `himalaya envelope list -a posteo ...`
- `himalaya account configure gmail`

## Verification

Validated locally against `himalaya v1.1.0`:

- `himalaya envelope list -a posteo --folder Archives.FOSS --page 2 --help`
- `himalaya account configure gmail --help`

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 </div>
 
 ```
-himalaya envelope list --account posteo --folder Archives.FOSS --page 2
+himalaya envelope list -a posteo --folder Archives.FOSS --page 2
 ```
 
 ![screenshot](./screenshot.jpeg)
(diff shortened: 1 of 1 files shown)


soywod:
The v2 is on master, and too many things changed in the doc. I manually apply your patch. Thanks for your contribution!

PR #663, opened 2026-04-02
fix(message): set Draft+Seen flags when saving to Drafts folder

### LINKED ISSUE #662: message save and template save do not set \Draft flag when saving to Drafts folder
## Environment

- Himalaya version: v1.2.0
- OS: Linux (Ubuntu 24.04)
- Backend: IMAP (tested with DavMail bridge to Exchange/O365)

## Description

When saving a message to the Drafts folder using `message save` or `template save`, the saved message does not have the `\Draft` IMAP flag. This causes IMAP clients (especially Outlook/OWA) to display the message as a regular sent message rather than an editable draft.

The interactive editor path (`PostEditChoice::RemoteDraft` in `pimalaya-tui/src/himalaya/editor.rs:143–147`) correctly sets both `Flag::Seen` and `Flag::Draft` via `add_message_with_flags`. The CLI save commands use `add_message` without flags.

## Steps to reproduce

```bash
echo "From: me@example.com
To: someone@example.com
Subject: Test draft

Body" | himalaya message save --folder Drafts
```

**Expected:** Message has `\Seen` and `\Draft` flags, appears as an editable draft in IMAP clients.

**Actual:** Message has no flags, appears as a regular sent message.

## Note on post-save flag workaround

Adding the flag after save does **not** work reliably:

```bash
himalaya flag add --folder Drafts <id> draft seen
```

On DavMail/Exchange, `flag add` creates a **duplicate** message with mangled headers (`Re:` prepended to subject, raw headers rendered in body) instead of modifying the original in place. The original unflagged message remains. This means the only reliable fix is setting the flags at IMAP APPEND time, not as a separate STORE operation.

## Root cause

In `src/email/message/command/save.rs:75`:

```rust
let id = backend.add_message(folder, msg.as_bytes()).await?;
```

This calls `add_message()` which defaults to empty flags (`email-lib`'s `add_message` delegates to `add_message_with_flags(folder, msg, &Default::default())`).

The same issue exists in `src/email/message/template/command/save.rs`.

Compare with the working interactive editor path in `pimalaya-tui:src/himalaya/editor.rs:143–147`:

```rust
backend
    .add_message_with_flags(
        DRAFTS,
        &email,
        &Flags::from_iter([Flag::Seen, Flag::Draft]),
    )
    .await?;
```

## Suggested fix

When the target folder matches the account's drafts alias, `message save` and `template save` should use `add_message_with_flags` with `[Flag::Seen, Flag::Draft]`, matching the interactive editor behaviour.

yetty:
## Summary

- `message save` and `template save` now set `\Seen` + `\Draft`
  flags when the target folder is the Drafts folder
- Fixes duplicate `add_message` call in `template save` (saved
  the message twice)
- Matches existing behaviour in the interactive editor path
  (`pimalaya-tui/src/himalaya/editor.rs`)

## Test plan

- [x] `cargo test` passes (includes new unit test for folder detection)
- [x] `cargo clippy` passes
- [x] Manual test: `message save --folder Drafts` shows as
  proper draft with `[Seen, Draft]` flags (DavMail/Exchange)
- [x] Manual test: `template save --folder Drafts` creates
  exactly one message with Draft flag (no duplicate)
- [x] Manual test: saving to non-Drafts folder does not add
  Draft flag

Closes #662

### DIFF
--- a/src/email/message/command/save.rs
+++ b/src/email/message/command/save.rs
@@ -6,7 +6,13 @@ use std::{
 
 use clap::Parser;
 use color_eyre::Result;
-use email::{backend::feature::BackendFeatureSource, config::Config, envelope::SingleId};
+use email::{
+    backend::feature::BackendFeatureSource,
+    config::Config,
+    envelope::SingleId,
+    flag::{Flag, Flags},
+    folder::FolderKind,
+};
 use pimalaya_tui::{
     himalaya::backend::BackendBuilder,
     terminal::{cli::printer::Printer, config::TomlConfig as _},
@@ -72,7 +78,17 @@ impl MessageSaveCommand {
                 .join("\r\n")
         };
 
-        let id = backend.add_message(folder, msg.as_bytes()).await?;
+        let id = if FolderKind::matches_drafts(folder) {
+            backend
+                .add_message_with_flags(
+                    folder,
+                    msg.as_bytes(),
+                    &Flags::from_iter([Flag::Seen, Flag::Draft]),
+                )
+                .await?
+        } else {
+            backend.add_message(folder, msg.as_bytes()).await?
+        };
 
         printer.out(MessageAdded { folder, id })
     }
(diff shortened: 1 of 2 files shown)


soywod:
Since v2, users have free hands on the flags they want to add, it should not be an issue anymore. Thanks for your contribution!

PR #667, opened 2026-04-10
feat(message): add --yes flag to reply command for non-interactive sending

### LINKED ISSUE #666: Feature Request: Add `--no-edit` / `--yes` flag to skip editor in `message reply` for non-interactive use
## Summary

Currently, `himalaya message reply` always opens `$EDITOR` for interactive editing before sending, even when `--body` is provided. There is no way to send a reply non-interactively (e.g., from a script or automated agent).

## Problem

When building automated email workflows (e.g., auto-reply agents using `himalaya` as the backend), the reply flow is blocked by the mandatory editor invocation:

```rust
// src/email/message/command/reply.rs
editor::edit_tpl_with_editor(account_config, printer, &backend, tpl).await?;
```

Even when `--body <text>` is supplied as a CLI argument (which pre-fills the template body), the editor is still launched. This makes it **impossible** to send replies from non-interactive scripts or automation pipelines without hacks like using a fake `$EDITOR`.

## Current workaround

A workaround exists by combining `template reply` + `message send`, but it requires additional processing of the raw template output (stripping headers, normalising charset, etc.), which is fragile:

```bash
himalaya template reply <id> "<body text>" \
  | <fix headers / charset> \
  | himalaya message send
```

This is error-prone and should not be necessary for such a common use case.

## Proposed Solution

Add a `--yes` (or `--no-edit`) flag to `message reply` (and symmetrically to `message forward` and `message write`) that skips the editor and sends immediately when `--body` is provided:

```
himalaya message reply --yes <id> "Approved, thank you."
```

Internally, the flag would bypass `editor::edit_tpl_with_editor` and call send directly:

```rust
if self.yes {
    backend.send_message_then_save_to_sent(&account_config, tpl.into_bytes()?).await?;
} else {
    editor::edit_tpl_with_editor(account_config, printer, &backend, tpl).await?;
}
```

## Use Cases

- **Automated approval agents**: scripts that process incoming approval emails and send standardised replies based on rules
- **CI/CD pipelines**: sending notifications or status updates via email without user interaction
- **Cron jobs / daemons**: scheduled email responders

## Related

- `message forward` has the same issue
- `message write` (compose) would also benefit from a `--yes` flag for fully scripted composition

Any feedback on the preferred flag name (`--yes`, `--no-edit`, `--send-immediately`, etc.) is welcome.

kamusis:
Closes #666

## Problem

`himalaya message reply` always opens an interactive editor (`$EDITOR`), making it impossible to use in automated or scripted workflows (e.g. CI pipelines, email agents).

The existing workaround — `himalaya template reply <id> | himalaya message send` — does **not** produce a proper reply: the generated template lacks `In-Reply-To` and `References` headers, so recipients see a new thread instead of a reply.

## Solution

Add a `--yes` / `-y` flag to `message reply`. When provided together with `--body`, the editor is skipped and the message is compiled and sent immediately via `send_message_then_save_copy`, preserving all reply headers (`In-Reply-To`, `References`) that `to_reply_tpl_builder` correctly sets.

If `--yes` is given without `--body`, the command exits with a clear error:
```
--yes requires --body to be set
```

## Changes

- `src/email/message/arg/yes.rs` — new `MessageYesArg` struct with `--yes` / `-y` flag
- `src/email/message/arg/mod.rs` — export `yes` module
- `src/email/message/command/reply.rs` — add `yes` field to `MessageReplyCommand`; branch on `--yes` to send directly or fall back to the existing editor flow

## Usage

```bash
# Non-interactive reply — skips editor, correct threading headers
himalaya message reply --yes <id> "Approved."

# Interactive reply — unchanged behaviour
himalaya message reply <id>
```

## Testing

Tested locally against a real IMAP/SMTP mailbox (Tencent Exmail). Verified:

- `himalaya message reply --yes <id> "text"` sends immediately without opening the editor
- The sent message is correctly threaded — recipient sees it as a reply with proper `In-Reply-To` and `References` headers, not a new email
- `himalaya message reply --yes <id>` (without `--body`) exits with the expected error: `--yes requires --body to be set`
- `himalaya message reply <id>` (without `--yes`) still opens `$EDITOR` as before — no regression

### DIFF
--- a/src/email/message/arg/mod.rs
+++ b/src/email/message/arg/mod.rs
@@ -3,6 +3,7 @@ use clap::Parser;
 pub mod body;
 pub mod header;
 pub mod reply;
+pub mod yes;
 
 /// The raw message argument parser.
 #[derive(Debug, Parser)]
--- a/src/email/message/arg/yes.rs
+++ b/src/email/message/arg/yes.rs
@@ -1,0 +1,13 @@
+use clap::Parser;
+
+/// The skip-editor argument parser.
+#[derive(Debug, Parser)]
+pub struct MessageYesArg {
+    /// Skip the editor and send the message immediately.
+    ///
+    /// This argument requires --body to be set. When provided, the
+    /// message is compiled and sent directly without opening the
+    /// editor defined in $EDITOR.
+    #[arg(long, short = 'y')]
+    pub yes: bool,
+}
--- a/src/email/message/command/reply.rs
+++ b/src/email/message/command/reply.rs
@@ -1,8 +1,12 @@
 use std::sync::Arc;
 
 use clap::Parser;
-use color_eyre::{eyre::eyre, Result};
+use color_eyre::{
+    eyre::{bail, eyre},
+    Result,
+};
 use email::{backend::feature::BackendFeatureSource, config::Config, flag::Flag};
+use mml::MmlCompilerBuilder;
 use pimalaya_tui::{
     himalaya::{backend::BackendBuilder, editor},
     terminal::{cli::printer::Printer, config::TomlConfig as _},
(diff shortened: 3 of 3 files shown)


soywod:
This argument will not be needed anymore in the v2, since composition moved away. It is not possible to fully script saving or sending. Thanks for your contribution!

PR #678, opened 2026-05-06
Update config.sample.toml

a-menshchikov:
Fix `gpg.sign-cmd` example: `--detach-sign` flag shold be used instead of `--sign` for a correct PGP signature.


soywod:
Himalaya drop PGP support in the v2, it may come back in a future release (more work needs to be done lib side to accept back PGP). I will keep in mind your fix in case of. Thanks for your contribution!

PR #682, opened 2026-05-19
fix: prevent panic when message send receives invalid email

PhoenixCPH:
The command `himalaya message send` panics with "index out of bounds: the
len is 0 but the index is 0" in mail-parser when given email content that
cannot be parsed.

Root cause:

  email-lib's SmtpContext::send() passes raw bytes to
  mail_parser::MessageParser::parse(). When parsing fails, the code
  substitutes Message::default(), which has an empty "parts" vector.
  The subsequent into_smtp_msg() calls msg.headers(), which internally
  calls msg.root_part(), which accesses self.parts[0] — panicking on
  the empty vector.

Fix:

  Add mail-parser as an explicit dependency and pre-validate the raw
  message in MessageSendCommand::execute() before handing it off to
  email-lib. If parsing fails, return a clean error with a hint about
  how to use the command correctly (pipe a valid .eml file).

This is a defensive fix at the himalaya level. The underlying issue in
email-lib (using Default::default() when parsing fails, then calling
methods that assume a non-empty message) could also be fixed there, but
this change prevents the panic regardless of which email-lib version is
used.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1853,6 +1853,7 @@ dependencies = [
  "clap_mangen",
  "color-eyre",
  "email-lib",
+ "mail-parser",
  "mml-lib",
  "once_cell",
  "open",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ clap_complete = "4.4"
 clap_mangen = "0.2"
 color-eyre = "0.6"
 email-lib = { version = "0.27", default-features = false, features = ["tokio-rustls", "derive", "thread"] }
+mail-parser = "0.9"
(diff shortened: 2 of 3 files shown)


PhoenixCPH:
QQ邮箱这么大的厂连个官方CLI都没有，还得靠咱们第三方客户端自己适配，难顶

soywod:
The v2 is on master and too many things changed. In the v2, the email lib parses headers and returns an error in case of failure, which should catch most of invalid mails:

https://github.com/pimalaya/io-email/blob/7ad5d05e801319c264c3745991c1fbef6d59e77d/src/smtp/message_send.rs#L68-L70

Thanks for your contribution!

PR #687, opened 2026-05-23
refactor: split composer config into subparts compose, reply and forward

### LINKED ISSUE #686: The `reply-with` and `forward-with` should support dedicated composer configurations
Currently, both the `message reply-with` and `message forward-with` commands pipe target messages into the generic `message.composer.*` configuration. 

For example, the current configuration only allows:

```toml
[message.composer.mml]
command = "mml compose"
```

We should extend the configuration to support command-specific composers, such as:

```toml
[message.composer.reply.mml]
command = "mml reply"

[message.composer.forward.mml]
command = "mml forward"
```

Bowen951209:
Fix #686.

Config extended from:

(Old)
```toml
[message.composer.mml]
command = "mml compose"
```

, where `compose-with`, `reply-with`, and `forward-with` all share the same composer command, to:

(New)
```toml
[message.composer.mml]
default = true
compose = "mml compose"
reply = "mml reply"
forward = "mml forward"
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -339,23 +339,6 @@ pub struct ComposerConfig {
     pub default: bool,
 }
 
-#[derive(Clone, Copy)]
-pub enum Composer {
-    Compose,
-    Reply,
-    Forward,
-}
-
-impl ComposerConfig {
-    pub fn get_command(&self, composer: Composer) -> &String {
-        match composer {
-            Composer::Compose => &self.compose_command,
-            Composer::Reply => &self.reply_command,
-            Composer::Forward => &self.forward_command,
-        }
-    }
-}
-
 /// Single reader entry under `[message.reader.<name>]`.
 #[derive(Clone, Debug, Deserialize, Serialize)]
 #[serde(rename_all = "kebab-case", deny_unknown_fields)]
--- a/src/config.rs
+++ b/src/config.rs
@@ -300,17 +300,35 @@ pub struct MessageConfig {
 }
 
 /// Single composer entry under `[message.composer.<name>]`.
+///
+/// For all shell command strings defined below:
+/// - The command is invoked via `sh -c`.
+/// - stdin behavior varies by command as documented below.
+/// - stdout is captured as the MIME draft.
+/// - stderr is inherited so the composer can prompt the user.
 #[derive(Clone, Debug, Deserialize, Serialize)]
 #[serde(rename_all = "kebab-case", deny_unknown_fields)]
 pub struct ComposerConfig {
-    /// Shell command line invoked via `sh -c`. Stdin carries the
-    /// source MIME bytes (empty for new messages); stdout is
-    /// captured as the MIME draft; stderr is inherited so the
-    /// composer can prompt the user.
+    /// Command used to write a brand new message.
+    ///
+    /// This is invoked by the `compose-with` and `mailto` commands.
+    ///
+    /// - When invoked by `compose-with`, stdin is empty.
+    /// - When invoked by `mailto`, stdin is piped with a pre-filled RFC 5322
+    ///   draft skeleton built from the parsed RFC 6068 `mailto:` URI parameters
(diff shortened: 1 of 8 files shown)


src/config.rs
```diff
@@ -300,14 +300,36 @@ pub struct MessageConfig {
...
+    ///
+    /// This is invoked by the `reply-with` command. The original message's
+    /// MIME bytes are passed via stdin.
+    pub reply_command: String,
+
+    /// Command used to forward an existing message.
+    ///
+    /// This is invoked by the `forward-with` command. The original message's
+    /// MIME bytes are passed via stdin.
+    pub forward_command: String,
```
soywod:
```suggestion
    #[serde(with = "command")]
    pub forward: Command,
```

src/config.rs
```diff
@@ -300,14 +300,36 @@ pub struct MessageConfig {
...
+    /// - When invoked by `mailto`, stdin is piped with a pre-filled RFC 5322
+    ///   draft skeleton built from the parsed RFC 6068 `mailto:` URI parameters
+    ///   (such as to, cc, bcc, subject, and body).
+    pub compose_command: String,
+
+    /// Command used to reply to an existing message.
+    ///
+    /// This is invoked by the `reply-with` command. The original message's
+    /// MIME bytes are passed via stdin.
+    pub reply_command: String,
```
soywod:
```suggestion
    #[serde(with = "command")]
    pub reply: Command,
```

src/config.rs
```diff
@@ -300,14 +300,36 @@ pub struct MessageConfig {
...
-    pub command: String,
+    /// Command used to write a brand new message.
+    ///
+    /// This is invoked by the `compose-with` and `mailto` commands.
+    ///
+    /// - When invoked by `compose-with`, stdin is empty.
+    /// - When invoked by `mailto`, stdin is piped with a pre-filled RFC 5322
+    ///   draft skeleton built from the parsed RFC 6068 `mailto:` URI parameters
+    ///   (such as to, cc, bcc, subject, and body).
+    pub compose_command: String,
```
soywod: Few things there. First I reconsider `_command` suffix, since there is no alternative (only command) it maybe does not make sense to add it. Secondly, my bad, we should not use `String`, but `std::process::Command` instead, with the `command` serde module from `pimalaya_config::command`. See [pimalaya/config/secret](https://github.com/pimalaya/config/blob/5df70410bfecb5e046346b80864273f101bed473/src/secret.rs#L32).

```suggestion
    #[serde(with = "command")]
    pub compose: Command,
```

soywod (changes requested):
Thanks for the contribution. While reading to your PR, I realized that `resolve_composer` is way too complicated. As a direct consequence, this feature becomes even more complicated. Ideally, on `Account` we should have a simple method `get_composer` that return the `ComposerConfig` matching either the name or `default = true`. Then each command `compose-with`, `reply-with` and `forward-with` would just need to select the right command `.compose`, `.reply` and `.forward`. No need the resolver nor the Compose kind at all. But this issue was already there, if you feel like resolving it now I would really appreciate, otherwise it's also fine I will take care of it after your PR. I just asked for few changes on the config shape. Feel free to share your opinion!

Bowen951209:
> we should not use String, but std::process::Command instead

But `Command` does not implement `Clone`, we will have to remove the `Clone` derives then.

soywod:
> > we should not use String, but std::process::Command instead
> 
> But `Command` does not implement `Clone`, we will have to remove the `Clone` derives then.

Nice catch, indeed we will have to remove `Clone`. I think it's fine, since we can only run one command at a time, I tried to consume configs as much as possible. I would even be surprised if there is 0 clone of config in the actual code base.

Bowen951209:
Updated config demo in PR message: Remove `_command` suffixes.

soywod:
Awesome, thank you. It will give us a good starting point to iterate. I still see some improvement to do, thanks for initiating it!

PR #688, opened 2026-05-24
feat(imap): add config option for IMAP ID (RFC 2971)

### LINKED ISSUE #676: IMAP ID command not sent when backend.extensions.id.send_after_auth = true
## Description

When connecting to imap.163.com (NetEase), the server requires clients to send an IMAP ID command (RFC 2971) after authentication. Without it, SELECT commands fail with:

```
NO SELECT Unsafe Login. Please contact kefu@188.com for help
```

163.com official FAQ confirms this: https://help.mail.163.com/faqDetail.do?code=d7a5dc8471cd0c0e8b4b8f4f8e49998b374173cfe9171305fa1ce630d7f67ac2eda07326646e6eb0

## Config

```toml
[accounts.personal]
email = "test@163.com"
default = true

backend.type = "imap"
backend.host = "imap.163.com"
backend.port = 993
backend.encryption.type = "tls"
backend.login = "test@163.com"
backend.auth.type = "password"
backend.auth.raw = "***"
backend.extensions.id.send_after_auth = true
```

## Expected Behavior

After LOGIN, an ID command should be sent before SELECT:
```
C: ID ("name" "Himalaya" "version" "1.2.0" "vendor" "pimalaya")
S: * ID ("name" "Coremail" ...)
S: ID completed
```

## Actual Behavior

The ID command is NOT sent. LOGIN is immediately followed by SELECT:
```
LOGIN test@163.com ... -> OK LOGIN completed
SELECT INBOX -> NO SELECT Unsafe Login
```

The `send_after_auth = true` option has no effect. Debug/trace output confirms no ID command is sent between LOGIN and SELECT.

## Environment

- Himalaya v1.2.0 (macOS aarch64)
- imap.163.com (NetEase)
- Server advertises ID capability: `* CAPABILITY IMAP4rev1 XLIST SPECIAL-USE ID LITERAL+ ...`

## References

- RFC 2971: https://www.ietf.org/rfc/rfc2971.txt
- 163.com official FAQ confirms IMAP ID is required

EzFavorites:
Add an IMAP extensions configuration system under `imap.extensions.*`, starting with RFC 2971 ID command support.

## Problem
Some IMAP servers (e.g. NetEase 163) require clients to send an IMAP ID command (RFC 2971) after authentication, before they allow mailbox SELECT. Without it, commands fail with `NO SELECT Unsafe Login`.

## Solution

Add a new `[accounts.<name>.imap.extensions.id]` config block:

```toml
[accounts.myaccount.imap.extensions.id]
send-after-auth = true
# optional overrides (when omitted, sensible defaults are used):
# name = "MyClient"
# version = "1.0"
# vendor = "MyOrg"
# support-email = "support@myorg.com"
```

When `send-after-auth` is `true`, an IMAP ID command is sent after authentication. The field values are optional overrides; when omitted, compiled-in defaults (name = "Himalaya", version = `CARGO_PKG_VERSION`, vendor = "Pimalaya", support-email = "pimalaya.org@posteo.net") are used.

## Backward Compatibility

`send-after-auth` defaults to `false`, preserving the previous behavior for all existing configurations.

## Implementation

- `src/config.rs`: Added `ImapExtensionsConfig` and `ImapIdConfig` structs, `extensions` field on `ImapConfig`
- `src/imap/client.rs`: Config-driven IMAP ID in `ImapClient::new()`
- `src/shared/client.rs`: Config-driven IMAP ID in `EmailClient` (shared subcommands path)
- `src/wizard/account.rs`: Initialize `extensions` with defaults

Closes #676

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -418,6 +418,68 @@ pub struct ImapConfig {
     /// to advertise the ANONYMOUS mechanism explicitly, set
     /// `sasl.anonymous = {}`.
     pub sasl: Option<SaslConfig>,
+
+    /// IMAP extensions configuration (RFC 2971 ID, etc.).
+    #[serde(default)]
+    pub extensions: ImapExtensionsConfig,
+}
+
+/// IMAP extensions configuration.
+///
+/// Controls optional IMAP extensions such as RFC 2971 (ID command).
+/// Each extension is nested under its RFC or feature name.
+///
+/// # Example
+///
+/// ```toml
+/// [accounts.myaccount.imap.extensions.id]
+/// send-after-auth = true
+/// name = "MyClient"           # optional override
+/// version = "1.0"             # optional override
+/// ```
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case", deny_unknown_fields)]
+pub struct ImapExtensionsConfig {
+    /// RFC 2971 ID command configuration. When `send-after-auth`
+    /// is true, an IMAP ID command is sent after authentication.
+    /// The field values are optional overrides; when omitted,
+    /// sensible defaults (name = "Himalaya", vendor = "Pimalaya",
+    /// version = `CARGO_PKG_VERSION`, support-email =
+    /// "pimalaya.org@posteo.net") are used.
+    #[serde(default)]
+    pub id: ImapIdConfig,
+}
+
+/// RFC 2971 IMAP ID command configuration.
+///
+/// Controls whether and how the IMAP ID command is sent after
+/// authentication. Some servers (e.g. NetEase 163) require this
+/// before allowing mailbox SELECT.
+///
+/// When `send-after-auth` is `false` (the default), no ID command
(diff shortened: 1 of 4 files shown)


soywod:
Thanks for initiating it. Indeed the IMAP extensions has not been ported to v2. I believe the port should go to `io-email` (I'm already on it, should come soon), then Himalaya CLI just maps the value from the config. I let you know when it lands.

soywod:
Side question, I'm not sure about customization of ID data by end users (Himalaya users). IMO they should be fixed by Himalaya itself?

soywod:
> Implementations MAY send less information than they have available or no information at all. Such behavior may be useful to preserve user privacy.

So in fact we should let user enable or disable fields they want to send, but not let them "edit".

soywod:
Feature added back on `master`:

```toml
imap.id.auto = true

# omitted fields are not sent by default, so users need to manually set the fields they agree to send, but they cannot modify them
imap.id.fields.name = true
imap.id.fields.vendor = true
# …
```

PR #692, opened 2026-05-28
fix(nix): rename buildFeatures → withFeatures

xav-ie:
`default.nix` calls nixpkgs's himalaya `package.nix` with
`buildFeatures` / `buildNoDefaultFeatures`. nixpkgs renamed those args
to `withFeatures` / `withNoDefaultFeatures` and now prints a deprecation
warning every time the old names are used:

    evaluation warning: buildFeatures is deprecated in favour of
      withFeatures and will be removed in the next release

This PR just renames them. No behavior change — nixpkgs still accepts
both names today, this only silences the warning ahead of the eventual
removal.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -19,8 +19,8 @@ pimalaya.mkDefault (
       }:
       (pkgs.callPackage "${nixpkgs}/pkgs/by-name/hi/himalaya/package.nix" {
         inherit lib rustPlatform;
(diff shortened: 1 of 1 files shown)


soywod:
The changes have been reverted on `nixpkgs`, see on `master`: https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/hi/himalaya/package.nix

Issue #699, opened 2026-06-10
docs: agent/LLM-friendly quickstart (llms.txt) + verified Gmail recipes for drafts, archive search, and folder aliases


andrewfurman:
## Summary

Proposal: add an LLM-agent-friendly quickstart (an `llms.txt`, or an `AGENTS.md`-style doc — whatever fits the project's docs philosophy) plus a small set of provider-specific recipes (Gmail first, then Outlook/iCloud) covering the things newcomers and AI agents consistently have to discover by trial and error: special folder names, saving drafts, and searching the archive.

I'm a heavy Himalaya user driving it both by hand and through LLM coding agents. The agents (and I) eventually figured everything out, but each nuance below cost real experimentation time, and almost none of it is in the README or `config.sample.toml`. Since the README already has an AI disclosure section and links an agent skill file maintained outside the project, it seems in spirit to give agents (and humans) a canonical, maintained quick reference here instead.

Everything below was verified live against Gmail over IMAP with **himalaya v1.2.0** — offered as ready-made content to adopt, trim, or rewrite.

## What's missing today (with references)

1. **Saving a draft** — there's no worked recipe for "save this RFC 5322 file as a draft" (asked before in discussion #258; the `\Draft` flag issue #662 is fixed, but the recipe itself is still undocumented):
   ```sh
   himalaya message save --folder "[Gmail]/Drafts" < message.eml
   # → Message <UID> successfully saved to [Gmail]/Drafts
   ```
   And the distinction: `message save` stores bytes as-is; `template save` compiles an MML template to MIME first.

2. **Gmail special folder names** — `config.sample.toml` shows aliases for sent/drafts/trash, but `"[Gmail]/All Mail"` (the archive, `\All`) is documented nowhere, and it's the answer to "how do I search all my mail":
   ```sh
   himalaya envelope list --folder "[Gmail]/All Mail" 'subject invoice and after 2026-01-01'
   ```
   Worth stating: all `[Gmail]/...` names need shell quoting, and every Gmail label appears as a top-level folder.

3. **Search syntax discoverability** — the v1.2.0 README documents no query syntax at all (master has one paragraph). Useful facts that are currently only learnable from `--help` + experimentation:
   - The query is Himalaya's own DSL (`from` / `to` / `subject` / `body` / `date` / `before` / `after` / `flag`, with `and` / `or` / `not` and `order by date desc`), compiled to server-side IMAP `UID SEARCH`.
[...]

PR #701, opened 2026-06-10
fix(imap): Gmail-compatible search fallback + docs from #699

### LINKED ISSUE #698: IMAP: Fallback to SEARCH when SORT not available
This is using Himalaya v2.

Listing envelopes, reading and sending messages works. However, search does not:

```sh
$ himalaya envelope search after "2026-06-09"
Error: IMAP SORT BAD error: Unknown command: UID SORT
```

This is with Gmail. 

The trace goes through handshakes correctly but ends in:

```
[TRACE io_imap::send] read line fragment: 2.0.l9ZV4U0d OK [READ-WRITE] INBOX selected. (Success)\r\n
[TRACE io_imap::send] command to write: 3.0.NMwD6SvY UID SORT (REVERSE DATE) UTF-8 SENTSINCE \"10-Jun-2026\"\r\n
[TRACE io_imap::send] read bytes: 3.0.NMwD6SvY BAD Unknown command: UID SORT\r\n
[TRACE io_imap::send] read line fragment: 3.0.NMwD6SvY BAD Unknown command: UID SORT\r\n
```

andrewfurman:
Follow-up to #699 and fix for #698 — this PR carries the two concrete outcomes of that discussion: the documentation improvements we agreed on (general docs for humans and agents alike, no agent-specific file), and a fix for the Gmail search breakage found while testing v2.

### 1. `fix(imap)`: UID SEARCH fallback when the server lacks SORT

The shared `envelope search` always issues `UID SORT`, even with no `order by` clause; Gmail (no RFC 5256 support) rejects it with `BAD Unknown command: UID SORT`, so **every shared search fails against Gmail** while `envelope list` works.

This PR detects the tagged NO/BAD refusal and reruns the same filter as a plain `UID SEARCH` on the same session:

- **without `order by`** — matching UIDs come back newest-first (descending UID ≈ arrival order, a close approximation of the `REVERSE DATE` default) and only the requested page is fetched;
- **with `order by`** — every matching envelope is fetched in chunks of 500 and sorted client-side before pagination (what v1 did; can be slow on huge match sets, documented as such).

Design notes, very open to direction here:

- The FETCH→`Envelope` conversion mirrors io-email's private `envelope_from` so fallback results are byte-identical to the SORT path's. The durable home for all of this is probably io-email's search coroutine itself (try SORT, degrade to SEARCH); I kept it CLI-side so himalaya works against Gmail without waiting on an io-email release, and the module doc marks it as deletable once io-email grows the fallback. Happy to port it down into io-email instead if you'd rather fix it there first.
- Error-driven rather than capability-driven detection (a server advertising SORT but failing it still gets the fallback; transport errors bubble up unchanged).

Verified live against imap.gmail.com on INBOX and `[Gmail]/All Mail`: filter searches, `order by date desc` / `order by subject asc`, pagination (`-p`/`-s`), `--json` output, and clean stderr. `cargo fmt` and `clippy` pass.

### 2. `docs`: README + config.sample.toml

- **Singular command names** — the shared-API examples still used the pre-3a1a981b plural forms (`mailboxes list`, `envelopes search`, `messages send`), which the binary rejects. Fixed throughout, leaving the genuinely-plural protocol groups (`jmap mailboxes`, `smtp messages`) untouched.
- **Search docs** — point to `envelope search --help` as the source of truth for the DSL (per your preference for `--help` over duplicated docs), note that provider-native operators (Gmail `X-GM-RAW`) are unsupported, and describe the new SORT-less fallback behaviour.
- **Draft/send recipe** — `message add -m drafts --flag draft < message.eml` and `message send --save sent`, in the composing section.
- **Reading messages section** — it was already in the TOC but missing from the body; now covers side-effect-free reads (`BODY.PEEK`, no `\Seen`), per-mailbox ids vs the stable `message-id`, and the `--json` envelope shape.
- **config.sample.toml Gmail notes** — labels as top-level mailboxes, the quoted `[Gmail]/` prefix, an `archive = "[Gmail]/All Mail"` alias example, and an app-password pointer next to SASL PLAIN.

Split into three conventional commits (1 fix, 2 docs) so anything you'd rather not take is easy to drop. One correction to my issue comment: the FAQ's `passwd` vs the sample's `password` is not a mismatch — both work via serde alias.

Fixes #698.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,17 +192,19 @@ Accounts can be (re)configured later with `himalaya account configure <name>`. T
 Backend-agnostic commands operate on the account's first configured backend, or the one selected with `-b/--backend`:
 
 ```
-himalaya mailboxes list
-himalaya envelopes list -m INBOX --page 2
-himalaya envelopes search from alice and after 2026-01-01 order by date desc
-himalaya flags add -m INBOX --flag seen 1:3,5
-himalaya messages copy --from INBOX --to Archives 42
-himalaya attachments download -m INBOX 42
+himalaya mailbox list
+himalaya envelope list -m INBOX --page 2
+himalaya envelope search from alice and after 2026-01-01 order by date desc
+himalaya flag add -m INBOX --flag seen 1:3,5
+himalaya message copy --from INBOX --to Archives 42
+himalaya attachment download -m INBOX 42
 ```
 
-When the `inbox` alias is configured under `[mailbox.alias]`, `-m/--mailbox` becomes optional: shared commands fall back to that id. With `[mailbox.alias] inbox = "INBOX"`, the calls above shorten to `envelopes list --page 2`, `flags add --flag seen 1:3,5`, etc.
+When the `inbox` alias is configured under `[mailbox.alias]`, `-m/--mailbox` becomes optional: shared commands fall back to that id. With `[mailbox.alias] inbox = "INBOX"`, the calls above shorten to `envelope list --page 2`, `flag add --flag seen 1:3,5`, etc.
 
-`envelopes list` is plain pagination, ordered by date descending. To filter or sort, use `envelopes search` with a trailing query covering `date`, `after`, `from`, `to`, `subject`, `body`, `flag` conditions (combined with `and`, `or`, `not`, grouped with parens) and an `order by date|from|to|subject [asc|desc]` sort chain. Date clauses target the `Date:` header (sent-at) on every backend.
+`envelope list` is plain pagination, ordered by date descending. To filter or sort, use `envelope search` with a trailing query covering `date`, `after`, `from`, `to`, `subject`, `body`, `flag` conditions (combined with `and`, `or`, `not`, grouped with parens) and an `order by date|from|to|subject [asc|desc]` sort chain. Date clauses target the `Date:` header (sent-at) on every backend. The full grammar lives in `himalaya envelope search --help`, which is the source of truth for the query DSL.
+
+The query DSL is himalaya's own and compiles to each backend's native search: provider-specific operators (Gmail's `in:`/`label:` syntax, `X-GM-RAW`, …) are not supported. On IMAP the search runs server-side as `UID SORT`; servers that lack the `SORT` capability (notably Gmail) are detected automatically and himalaya degrades to a plain `UID SEARCH` — without `order by` the results come back newest-first (arrival order, page fetched only), with `order by` every match is fetched and sorted client-side, which can be slow on very large mailboxes.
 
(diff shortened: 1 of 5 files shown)


soywod:
Thanks a lot for the contribution, it sets up a good base we can iterate on.

First of all I believe it would be better to split this PR into smaller ones. Should be easy to do since all is already scoped by commit. It will make review easier.

Then I will give my first feeling about each scope:

- [8387542](https://github.com/pimalaya/himalaya/pull/701/changes/83875429d399b92a53b73b3374cdb9c953b31d62): I'm afraid this is not sustainable, because every provider has its own quirks. We will end up with a config sample full of annotiations here and there. In the v1.2.0 we had a [configuration per provider](https://github.com/pimalaya/himalaya/tree/v1.2.0#configuration) section, and I think it was a good starting point. OAuth 2.0 made sections quite verbose, but since it moved away from Himalaya we can definitely put them back.

- [6af1f47](https://github.com/pimalaya/himalaya/pull/701/changes/6af1f476b6b973a83ee5d189e3bc989cd3fab990): looks great. I think I should define a rule for all Pimalaya projects: always singular for public API (command, subcommand, config) with plural hidden aliases, but plural inside the code (struct) when it contains one or more items. A good example is mailbox aliases: should be `mailbox.alias` in config with `aliases` as alternative, but `pub aliases: HashMap<String, String>`. Which requires a `#[serde(default, rename = "alias", alias = "aliases")]`. If you feel like renaming inside the PR go ahead, but I can definitely do it by myself another time. Most important is the documentation.

- [c649818](https://github.com/pimalaya/himalaya/pull/701/changes/c649818538c79b3c2fda19f92669de78c95e819e): I agree with the algorithm (which is similar to the v1), but not with the architecture nor the structure. Shared mod should not contain protocol-specific code, so a `pub mod search_fallback;` does not have its place behind IMAP cargo feature. The fallback should not be executed when the SORT fails: I would rather check the capability or take user's choice. Last, we should put as much logic in coroutines as possible. In io-imap we have search coroutine, sort coroutine, so it's a perfect occasion to have a sort with search coroutine fallback depending on an a given option. The manual local sort algorithm should be there. This way, io-email uses this new coroutine as default search, and everyone benefits from the fallback. Not sure how clear I am, but anyway I needed to fix other bugs in io-imap, I can take care of it as well.

I close so you can split and decide wherever you want to tackle this sort or search fallback inside io-imap or not.

andrewfurman:
Thanks for the detailed feedback! Split as requested: docs in #703 (naming rule applied), provider configuration section in #704. Happy to leave the sort/search fallback coroutine in io-imap to you.

PR #704, opened 2026-06-11
docs: add Gmail section under Configuration (from #701)

andrewfurman:
Replaces the config.sample.toml Gmail annotations from #701 with a per-provider section in the README, modeled on the v1.2.0 Configuration section. Starts with Gmail only; other providers can follow the same pattern.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@
   - [Nix](#nix)
   - [Sources](#sources)
 - [Configuration](#configuration)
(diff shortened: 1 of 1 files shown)


soywod (approved):
Looks good to me, we will need to import back other providers in this section. Thank you!

andrewfurman:
Other providers imported back in #705.

PR #705, opened 2026-06-11
docs: import remaining provider sections back from v1.2.0

andrewfurman:
Follow-up to #704: imports the Proton Mail, Outlook and iCloud Mail sections back from the v1.2.0 README, translated to the v2 config syntax. All four example configs parse against the current binary.


soywod (approved):

soywod (approved):
Thanks!

PR #728, opened 2026-08-04
fix(message): parse --from as a real mailbox

### LINKED ISSUE #727: --from "Name <email>" breaks the composed From header and sending
Passing a display name to `--from` produces a broken message:

```
himalaya message compose -a myaccount --from "Name <me@example.com>" --to test@example.com --subject "test" --body "hej"
```

The resulting header is:

```
From: <Name <me@example.com>>
```

nested angle brackets. Sending it fails with:

```
Error: SMTP MAIL FROM failed: rejected 501 5.1.7 Bad sender address syntax
```

A bare address without a display name works fine and sends, just without a name shown to the recipient.

Version: himalaya 2.0.0 (stable)

truls27a:
--from went straight into mail_builder's plain-string Address conversion, which doesn't parse "Name <addr>" and stores the whole string as the address. This produced `From: <Name <addr>>` (nested angle brackets) and failed to send with an SMTP "Bad sender address syntax" error.

--from is now parsed with mail_parser first, so both bare addresses and "Name <addr>" pairs produce a correct header.

Before:
```
$ himalaya message compose --from "Name <me@example.com>" ...
From: <Name <me@example.com>>
```
fails to send with `SMTP MAIL FROM failed: rejected 501 5.1.7 Bad sender address syntax`

After:
```
From: "Name" <me@example.com>
```
sends fine.

Fixes #727

### DIFF
--- a/src/shared/message/builder.rs
+++ b/src/shared/message/builder.rs
@@ -71,7 +71,7 @@ pub fn build(args: BuilderArgs<'_>, source: Option<SourceArgs<'_>>) -> Result<Ve
     let mut builder = MessageBuilder::new();
 
     if let Some(from) = args.from {
-        builder = builder.from(from);
+        builder = builder.from(parse_mailbox(from)?);
     }
     if !args.to.is_empty() {
         builder = builder.to(addresses(args.to));
@@ -160,6 +160,31 @@ pub fn build(args: BuilderArgs<'_>, source: Option<SourceArgs<'_>>) -> Result<Ve
     builder
         .write_to_vec()
         .map_err(|err| anyhow!("serialize composed message: {err}"))
+}
+
+fn parse_mailbox(value: &str) -> Result<Address<'static>> {
+    let synthetic = format!("From: {value}\r\n\r\n");
+    let parsed = MessageParser::new()
+        .parse(synthetic.as_bytes())
+        .ok_or_else(|| anyhow!("Could not parse address `{value}`"))?;
+
+    let mailbox = match parsed.from() {
+        Some(mail_parser::Address::List(list)) => list.first(),
+        Some(mail_parser::Address::Group(groups)) => {
+            groups.first().and_then(|g| g.addresses.first())
+        }
+        None => None,
+    }
+    .ok_or_else(|| anyhow!("Could not parse address `{value}`"))?;
+
+    let email = mailbox
+        .address
+        .clone()
+        .ok_or_else(|| anyhow!("Address `{value}` has no email"))?
+        .into_owned();
+    let name = mailbox.name.clone().map(|s| s.into_owned());
+
+    Ok(Address::new_address(name, email))
 }
 
 fn addresses(values: &[String]) -> Address<'static> {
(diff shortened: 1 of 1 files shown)


src/shared/message/builder.rs
```diff
@@ -162,6 +162,31 @@ pub fn build(args: BuilderArgs<'_>, source: Option<SourceArgs<'_>>) -> Result<Ve
         .map_err(|err| anyhow!("serialize composed message: {err}"))
 }
 
+fn parse_mailbox(value: &str) -> Result<Address<'static>> {
+    let synthetic = format!("From: {value}\r\n\r\n");
+    let parsed = MessageParser::new()
```
soywod, line 167: If I don't mistake, there is a way to parse only the header instead of parsing a whole fake message, sth like `mail_parser::parsers::MessageStream::new(bytes).parse_address()`.

soywod (changes requested):
Thanks a lot for the fix. The direction seems good to me, except the method used (see my comments). Then we can merge.

bobberb:
`MessageStream::parse_address` does work here, with one catch worth saving you a round trip: it flushes its last token on the terminating newline, so a bare `truls@example.com` returns nothing without one. `Name <addr>` happens to work either way, since the `>` closes the token. Appending `\n` fixes it.

```rust
use mail_parser::{HeaderValue, MessageParser, parsers::MessageStream};

fn parse_mailbox(value: &str) -> Result<Address<'static>> {
    // NOTE: the header-value parser flushes its last token on the
    // terminating newline, which a bare address would not otherwise have.
    let value_nl = format!("{value}\n");
    let parsed = MessageStream::new(value_nl.as_bytes()).parse_address();

    let mailbox = match &parsed {
        HeaderValue::Address(mail_parser::Address::List(list)) => list.first(),
        HeaderValue::Address(mail_parser::Address::Group(groups)) => {
            groups.first().and_then(|group| group.addresses.first())
        }
        _ => None,
    }
    .ok_or_else(|| anyhow!("Could not parse address `{value}`"))?;

    let email = mailbox
        .address
        .as_ref()
        .ok_or_else(|| anyhow!("Address `{value}` has no email"))?
        .to_string();
    let name = mailbox.name.as_ref().map(|name| name.to_string());

    Ok(Address::new_address(name, email))
}
```

I ran both versions side by side on `truls@example.com`, `Truls Borgvall <truls@example.com>`, `"Borgvall, Truls" <truls@example.com>`, `=?utf-8?B?VHJ1bHM=?= <truls@example.com>`, a padded address, `a@e.com, b@e.com`, `Truls <truls@example.com> (comment)`, `not-an-address` and `""`. Identical on all nine, encoded-word decoding included. Your two tests pass unchanged, as does the rest of the suite; fmt and clippy clean.

soywod:
I wanted this issue to be part of the next release, so I allowed myself to merge your modified PR on `master` directly. I put you as co-author. Thanks for your help!

truls27a:
Ah I just got done updating the PR when I saw your message hahahaha. No worries, thank you for taking your time and implementing this! :)

soywod:
> Ah I just got done updating the PR when I saw your message hahahaha. No worries, thank you for taking your time and implementing this! :)

Sorry for the inconvenience, I should have asked before. The new release will come tonight, stay tuned ;)

bobberb:
"I just want to live in a world of sane tooling for 40 year old tech - is that too much to ask?"

Soywod got tired of asking. King 👑.

Issue #726, opened 2026-08-04
message compose/reply/forward don't set a From header without --from


truls27a:
Composing a message without passing `--from` produces no `From:` header at all:

```
himalaya message compose -a myaccount --to test@example.com --subject "test" --body "hej"
```

Output has `To`, `Subject`, `Message-ID`, `Date`, and MIME headers, but no `From`. Piping it into `message send` fails with:

```
Error: No `From:` header found in raw message
```

`display-name` and `email` are set in the account config, and `account check` passes fine for both imap and smtp, so himalaya is reading the config. In v1, `template reply` used those fields to fill in `From:` automatically. In 2.0.0 `--from` seems to be required on every single compose/reply/forward call.

Version: himalaya 2.0.0 (stable)

soywod:
Oups sorry, I thought it was the original issue from #733, reopening.

soywod:
Indeed, since composition moved away in v2, display-name and email have been removed from config. But, to be fair, the `message compose` is a composer. Lightweight and verbose, but it's a small composer. Himalaya TUI reuses the same config file as the CLI, and it requires display-name + email (because it's a more complete tool, it requires composition inside). So I would not be against putting them back and auto-fill the `message compose --from` header. Feel free to propose a PR.

soywod:
Field `display-name`, `email`, `signature` and `signature-delim` came back to the config, on `master`. Thanks for the report!

truls27a:
Nice, thank you so much for implementing this! :)

PR #733, opened 2026-08-15
fix(send): inject Date header when missing on send path

AnsenIO:
## Symptom

Sending a raw RFC 5322 message that has no `Date:` header delivers it without one. The receiving IMAP server then files it at the Unix epoch — the message shows up dated **1970-01-01**.

Reproduced twice on 2026-08-15 with himalaya v1.2.0 (linux musl aarch64):

```console
$ printf 'From: me@example.com\r\nTo: you@example.net\r\nSubject: date test\r\n\r\nbody\r\n' > /tmp/no-date.eml
$ himalaya template send < /tmp/no-date.eml
# delivered, but arrives in the recipient's mailbox stamped 1970-01-01
```

On master (v2.0.0) the equivalent entry point is `himalaya message send < /tmp/no-date.eml`, which has the same gap.

## Root cause

The shared send path never inspects the message it transmits:

- `message send` resolves raw bytes via `MessageArg` and passes them through `handler::route` → `handler::apply` → `EmailClient::send_message` untouched.
- The built-in composers (`compose` / `reply` / `forward`, `shared/message/builder.rs`) assemble MIME with `mail_builder` and never set a `Date:` either, so the gap affects every send path, not just raw input.
- None of the backend adapters (`smtp/backend.rs`, `jmap/backend.rs`, `gmail/backend.rs`, `msgraph/backend.rs`) inject one; each transmits the bytes it is given.

RFC 5322 §3.6 lists the origination date as one of the only two mandatory header fields (with `From:`).

## Fix

`EmailClient::send_message` (`src/shared/client.rs`) is the single choke point every shared send path funnels through. It now prepends `Date: <now>` — RFC 5322 date-time with the local UTC offset via `chrono::Local` (`chrono` with `clock` is already a dependency) — when the message carries no `Date:` header. A message that already has one passes through **byte-identical**; the header is never overwritten. The injected line matches the message's own line-ending convention (CRLF or LF).

The `smtp send` plumbing command (explicit-envelope raw transaction) is intentionally left byte-exact.

## Test

Three unit tests next to the fix in `src/shared/client.rs`:

- `send_injects_date_header_when_missing` — no `Date:` in → valid, parseable `Date:` present after the transform; original bytes preserved after the prepended line
- `send_preserves_existing_date_header_byte_identically` — existing `Date:` → output equals input exactly
- `send_injected_date_matches_lf_line_endings` — an LF-only message gains an LF-terminated `Date:` line

TEST_RESULT_PLACEHOLDER

## Notes

- I searched open+closed upstream issues for "Date header", "missing date", "1970" and "epoch" before filing — nothing matched (#726 is about `From:`, unrelated), so no issue is duplicated here.
- Per this repo's Cairn convention the PR includes `cairn/changes/send-missing-date-header/` (`status: active`); folding the delta into `cairn/spec/commands.md` and writing the log entry are left for landing, as noted in its `tasks.md`.
- `Message-ID:` injection (also recommended by RFC 5322 §3.6) is intentionally out of scope to keep the change minimal.

### DIFF
--- a/cairn/changes/send-missing-date-header/delta.md
+++ b/cairn/changes/send-missing-date-header/delta.md
@@ -1,0 +1,12 @@
+# Delta
+
+## ADDED Requirements
+
+The shared send path SHALL ensure the message carries an origination
+date before handing it to any backend: `EmailClient::send_message`
+prepends `Date: <now>` (RFC 5322 date-time with the local UTC offset)
+when the message has no `Date:` header, and leaves a message that
+already carries one byte-identical. This covers every shared command
+that sends (`message send`, `compose`, `reply`, `forward`) across
+every send-capable backend (SMTP, JMAP, Gmail, Graph). The `smtp send`
+plumbing command stays byte-exact and is exempt.
--- a/cairn/changes/send-missing-date-header/proposal.md
+++ b/cairn/changes/send-missing-date-header/proposal.md
@@ -1,0 +1,45 @@
+---
+cairn: change
+id: send-missing-date-header
+status: active
+created: 2026-08-15
+---
+
+# Send missing Date header
+
+## Why
+
+A message sent through the shared send path without a `Date:` header
+is delivered as-is: nothing between the user's bytes and the wire adds
+one. Receiving IMAP servers then display the message at the Unix epoch
+(1970-01-01), which breaks sorting, threading and search for the
+recipient.
+
+Reproduced on v1.2.0 with `himalaya template send < message.eml`
+where `message.eml` carries no `Date:`; the delivered copy arrives
+stamped 1970-01-01. On master the same gap exists: `message send`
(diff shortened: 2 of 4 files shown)


AnsenIO:
Follow-up verification: full crate suite green with the patch — `cargo test -p himalaya` → 87 passed, 0 failed (includes the 3 new send_* tests).

soywod:
Thanks for the initiative. I will unfortunately decline. This issue is exactly the reason why Himalaya v2 got rid of composition: because it's a hard topic that requires its own attention. Himalaya does already enough.

In v1, composition was part of the CLI, this PR could have been merged.
The v2 expects a valid MIME message, the content is an opaque box. What you need is a top layer that composes properly messages. The v2 exposes a lightweight `message compose` that helps your composing a message, and it already set the Date. How do you actually compose message?

PR #735, opened 2026-08-16
feat(maildir): surface custom keywords on read

bobberb:
Moved from pimalaya/io-email#3. Smaller than that one: himalaya already delegates writes to the inner client, so only the config surface and the read path were missing.

`envelope list "flag NonJunk"` matches on IMAP and JMAP and silently matches nothing on Maildir, since only the six standard letters became flags. Both new keys default to off. Reading now goes through io-maildir instead of a local filename parse, which also drops a copy splitting the info section at the last comma rather than at `:2,`.

Read only: `FlagArg` is a closed enum, so nothing writes a custom keyword on any backend.

One caveat: on io-maildir 0.2.1 `flag add`/`flag remove` strip a keyword they never touched. That happens today already, unseen; showing keywords makes it visible. Pending as pimalaya/io-maildir#3, picked up on 0.2.2. Say so if you would rather ship only `keywords-header` until then.

Verified: fmt, clippy, 104 tests (11 new), the three reduced feature builds, and end to end against a throwaway Maildir. Cairn change, spec fold and log are in the commit.

AI disclosure per AI_POLICY.md: written with Claude Code, which did the exploration, edit, tests and prose. I set the design and the caveat call, and ran the gates.


soywod (changes requested):
Thanks, the direction is right and the config surface is the shape I want. Four things block it as written, all verifiable against io-maildir 0.2.1.

## 1. The `Fixed` entry is not true on 0.2.1

The claim is that the local copy split at the last comma "where io-maildir uses the `:2,` marker". On 0.2.1 both `MaildirFlags::with_dovecot`
(src/flag.rs:110) and `impl From<&MaildirFsPath>` (src/flag.rs:38) do `file_name.rsplit_once(',')`, character for character what `parse_filename_flags`
did. A `new/` name ending in `,S=1234` still yields a false `\Seen` after this PR.

The marker split only exists in 69f9939, merged yesterday and unreleased. So the CHANGELOG `Fixed` entry and the sentence added to cairn/spec/backends.md
describe behaviour the binary will not have. Drop them, or land after 0.2.2.

## 2. io-maildir#3 is merged, and the exposure is wider than described

Merged 2026-08-15; only the 0.2.2 tag is missing, so `io-maildir = "0.2"` still resolves to 0.2.1. There the entry locate parses through
`From<&MaildirFsPath>` and drops the slot letters before `rename_with_flags` (src/flag/add.rs:127). That is not only `flag add` and `flag remove`: `message
read` marks seen through the same `add_flags`, so with the option on, reading a message deletes its keywords. I will publish 0.2.2 and want the
requirement pinned to `0.2.2` here before this lands.

## 3. Residual loss on 0.2.2, undocumented

`store_flags` maps `FlagOp::Set` to `set_flags`, which replaces the whole set (upstream now documents this on the method). Since `FlagArg` cannot name a
keyword, `himalaya flag set seen` always wipes them. That belongs in config.sample.toml next to the other caveat.

## 4. The sidecar is per mailbox, not "at the Maildir root"

`MaildirDovecotLoad::new` joins `maildir.path()` with `dovecot-keywords` (src/dovecot/load.rs:56), and `keyword_table(&maildir)` loads it per resolved
mailbox, which is correct. The wording is wrong in config.sample.toml, the changelog, the log and the spec requirement, and that last one is normative
text.

## Layering

You cite the triage rule to drop the filename parse, then keep the keyword resolution here. io-maildir's client already owns `dovecot_keywords` and
`keywords_header` and honours them on store; this PR sets those fields and reads them back through `Deref` to hand-roll the read half. The symmetric answer
is a read-side API upstream, resolving an entry's flags against the client's own settings. Himalaya then keeps the config fields and the
`MaildirFlag::Keyword -> Flag::from_raw` arm, and backend.rs barely changes. I would rather have that, since the same read logic is needed by neverest and
the replica work.

## Smaller

- `#[derive(Default)]` on `MaildirConfig` makes `root: ""` constructible and adds `..Default::default()` at four sites. Naming both fields explicitly keeps
the invalid state unconstructible.
- `KeywordHeaderConfig` has no domain prefix, unlike `ImapIdConfig`. Either `MaildirKeywordHeaderConfig`, or nest the pair as `maildir.keywords.dovecot`
and `maildir.keywords.header`, which also beats two unrelated key spellings.
- `envelope_from_entry` and `flags_from_entry` thread the same `(table, header)` pair. That is a struct.
- tasks.md says "knobs"; use options. config.sample.toml backticks the dovecot-keywords filename, we do not backtick paths. Test name
`a_maildir_root_alone_keeps_keywords_off` has a stray leading `a_`.
- `extract_keywords_header` rescans raw bytes although `entry.parsed()` is already computed alongside.

## Good

Config mirror enum follows the alpn precedent so the schema stays backend-free. Sidecar load skipped when off and done once per listing. Tests are
behavioural rather than construction-only, and the ordering expectations match `Flag`'s manual `Ord`. The "no write surface" claim checks out, both
`FlagArg` enums are closed.

Note it conflicts with work in flight on CHANGELOG.md, cairn/spec/backends.md and src/maildir/backend.rs, the last one directly in `envelope_from_entry`.
Rebase once 0.2.2 is out.

--- Claude Opus 5

bobberb:
All four confirmed against 0.2.1, two of them my mistakes. Pushed 002f1ae for everything that does not need the tag.

1. Right, and it was careless: I checked `with_dovecot` existed in 0.2.1 without checking what it does, and it is `rsplit_once(',')` exactly like the local copy. The marker split is only in 69f9939. Dropped the `Fixed` entry and reverted the sentence in cairn/spec/backends.md to say the format has one definition, which is true today.

2. Also right, and worse than I wrote: `get_message` marks seen through the same `store_flags`, so `message read --seen` deletes keywords too. Said so in the proposal and the log.

3. Documented in config.sample.toml, the spec requirement and the changelog: reading is not a round trip, `flag set` replaces the whole set and drops what it cannot name.

4. Confirmed by fixture rather than by reading: root and a `Work` subfolder each with their own `dovecot-keywords` mapping slot 0 differently now list `RootKeyword` and `WorkKeyword` respectively. My original fixture had one mailbox so it could not have caught this. Fixed in all four places.

Smaller: options nested as `maildir.keywords.dovecot` / `maildir.keywords.header` (your second suggestion, since it does fix the two spellings), `MaildirKeywordHeaderConfig` prefixed, `Default` derive dropped and the four sites name the field, test renamed, "knobs" and the backticked path gone.

Left for the rebase, since they touch the code you want restructured: the `(table, header)` pair as a struct, and `extract_keywords_header` reusing `entry.parsed()`.

On layering, agreed, and the neverest argument settles it. I will send io-maildir a read-side API resolving an entry's flags against the client's own `dovecot_keywords` and `keywords_header`, so himalaya keeps the config plus the `Keyword -> Flag::from_raw` arm. One question so I do not collide with your release: do you want that in 0.2.2, or 0.2.2 now with 69f9939 alone and the read API in 0.2.3? I will pin the requirement to whichever carries it, and rebase then.

soywod:
Small conflict to fix and we can merge :)

soywod:
After reading more carefully, I split myself your PR into `io-maildir` and Himalaya. Some stuff were not belonging to the CLI. I put you as co-author, thanks for your work :)

bobberb:
Cheers

PR #746, opened 2026-09-08
gmail: support the shared envelope search

sjauv:
## Problem

The gmail backend bails on the shared search query, so `himalaya envelope search` — and every client built on it, from unread counts to MCP wrappers — fails outright on Gmail accounts with `Gmail does not support the shared envelope search`.

## Solution

Translate the filter half of the shared query into the `messages.list` `q` syntax, the same one the Gmail search box speaks:

- leaf conditions map to operators: `from alice` becomes `from:alice`, `body x` becomes a bare full-text term
- conjunctions flatten into whitespace, disjunctions into brace groups
- negations expand through De Morgan so a `-` only ever prefixes a leaf, the shape Gmail matches reliably
- exact dates bracket the day, `after:` / `before:` being day-granular bounds
- flags map to their `is:` operators, mirroring the existing label pairs in `gmail::backend`; `flag answered` has no Gmail operator and bails rather than returning wrong hits

The sort half has no server-side counterpart, `messages.list` keeping a fixed newest-first order, so it orders the fetched page client-side. The caveat — page boundaries stay in Gmail's default order — is noted in the module docs.

Microsoft Graph still bails; the same treatment should port over if wanted.

## Tests

11 new translator unit tests (quoting, De Morgan expansion, date bracketing, flag mapping, error cases); full suite green (121/121). Verified live against a Gmail account: `envelope search "not flag Seen"` and subject/from searches return results, and an MCP email toolset built on the shared search went from erroring to working.


soywod:
Thanks for your contribution!

